Issue #4, opened 2015-08-07
Use the new etcd client


victorcoder:
The current etcd client is being deprecated, the new client https://godoc.org/github.com/coreos/etcd/client will have all the focus in the future.

samgaw:
Would you consider Docker's libkv instead of the etcd client?

github.com/docker/libkv

victorcoder:
I thought this should be abstracted to allow several storage backends! thanks @samgaw I wanted this to exists.

Definitely worth considering.

From what I see in the docs the transition should be straightforward.

samgaw:
We moved to Consul so the etcd code was the first thing I skimmed through to see how baked in it was but it's pretty minimal.

If I get a chance over the next couple of days I'll RTFM libkv but I wanted to check first that there wasn't a specific reason for updating the etcd client.

victorcoder:
@samgaw Look at #21

PR #21, opened 2015-08-31
Libkv

victorcoder:
This should supersede #4, I consider it experimental, at least until libkv is not considered experimental too https://github.com/docker/libkv/issues/25

This will probably happen when etcd api v3 is ready


victorcoder:
Self argumentation: Being able to use different KV stores is for me, a great feature. It was a feature planned for the long term but with libkv has been almost trivial to implement.

On one side libkv is not completely stable nor dkron.

The plan is to fully support etcd while dkron is still not stable and the timeframe should match the timeframe etcd api v3 will be ready and also libkv will be stable.

Said that, I should be able to deal with any bug found in libkv to help them.

Test suite is passing and this should give enough confidence.

I'll be merging this and supporting any bug found with other stores.

PR #57, opened 2015-12-17
Remove not needed var

naxhh:
As you don't do anything else with this var. I think it's not needed.

Maybe you can move the condition to `a.isGraceful(sig)` for more redeable code. But that's your call :)


victorcoder:
I don't find this to be more readable, it's easier to read a positive disjunction than two negatives conjunction.

naxhh:
As I said. was your call :)

PR #73, opened 2016-02-10
Fixes node processing with wrong status

### LINKED ISSUE #71: Cluster missing executions
> As an administrator of a dkron cluster I want my users to have a reliable cron system which guarantees "at least once" execution of their cron tasks.

## My setup:

3 AWS EC2s running CoreOS stable, all three machines being part of an etcd cluster, all three machines running dkron 0.6.3 as a cluster. All machines does have `role: executor` tag and the following cron:

```json
{
	"name": "date",
	"schedule": "0 */5 * * * *",
	"command": "/usr/bin/date",
	"owner": "Knut Ahlers",
	"owner_email": "knut****.com",
	"run_as_user": "core",
	"success_count": 58,
	"error_count": 0,
	"last_success": "2016-02-05T10:45:00.00621402Z",
	"last_error": "0001-01-01T00:00:00Z",
	"disabled": false,
	"tags": {
		"role": "executor:1"
	}
}
```

## Expected result:

The cron is executed every 5 minutes and will succeed every single time as it just calls `/usr/bin/date` (which is the location of the `date` utility on CoreOS).

## Observed result:

The task is sent to one of the machines every 5 minutes and gets eventually executed. The cron is missing several executions having no end date, no running `date` utility on the machine and a `failed` state.

screen shot 2016-02-05 at 11 43 02

## Question

Can you help me with this / tell me what I'm doing wrong? Is there anything I can do to prevent this?

victorcoder:
Currently we're including in the node filtering, nodes that could
potentially be gone.

fixes #71

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -763,10 +763,12 @@ func (a *AgentCommand) processFilteredNodes(job *Job) ([]string, map[string]stri
 			}
 
 			for _, member := range a.serf.Members() {
-				for mtk, mtv := range member.Tags {
-					if mtk == jtk && mtv == tv {
-						if len(nodes) < count {
-							nodes = append(nodes, member.Name)
+				if member.Status == serf.StatusAlive {
+					for mtk, mtv := range member.Tags {
+						if mtk == jtk && mtv == tv {
+							if len(nodes) < count {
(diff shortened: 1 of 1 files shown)


Luzifer:
:+1:

PR #102, opened 2016-04-06
LibKV authentication

seiffert:
In order to run dkron with an etcd cluster that is hosted somewhere else, we would like to use etcd authentication. I've [added this feature to libkv](https://github.com/docker/libkv/pull/121) and would like to add these configuration options to dkron as well.

The other libkv backends will get the same options in the future: https://github.com/docker/libkv/issues/124

What do you think, @victorcoder?

### DIFF
--- a/dkron/rpc_test.go
+++ b/dkron/rpc_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestRPCExecutionDone(t *testing.T) {
-	store := NewStore("etcd", []string{etcdAddr}, nil, "dkron")
+	store := NewStore("etcd", []string{etcdAddr}, nil, "dkron", "", "")
 
 	// Cleanup everything
 	err := store.Client.DeleteTree("dkron")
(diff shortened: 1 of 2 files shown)


victorcoder:
I followed your work in libkv @seiffert and it's awesome.

Just that we'll need a test for this.

seiffert:
I agree :) I'll add one soon

victorcoder:
@seiffert  ❤️ to add this

seiffert:
I rebased the branch and will add a test now. Just to clarify: You mean a test for the configuration, right? So I would probably need to extend https://github.com/victorcoder/dkron/blob/f721d1c784c437d68cda286fe66a70d9ef60bd0e/dkron/agent_test.go#L331-L370, right?

seiffert:
Sorry, I'm no longer working on this because I don't use dkron at work anymore. Should I just close this PR or would you like to take over from here?

victorcoder:
Thanks @seiffert, yes I'll take care

PR #121, opened 2016-05-11
Return last 100 executions in dashboard and API

mlafeldt:
by fixing off-by-one error. At the moment, only 99 executions are ever shown.


mlafeldt:
You're welcome!

PR #125, opened 2016-05-13
WIP: Support long commands

### LINKED ISSUE #122: Support for long Job Commands
We are still struggling with the length limit for commands which was made explicit in #114. To solve this issue, we'd like to suggest the following change: Instead of [sending the whole execution via serf](https://github.com/victorcoder/dkron/blob/f721d1c784c437d68cda286fe66a70d9ef60bd0e/dkron/agent.go#L702-L715), we could just send the job's name and something identifying (maybe [the `Group`](https://github.com/victorcoder/dkron/blob/f721d1c784c437d68cda286fe66a70d9ef60bd0e/dkron/agent.go#L704)?) via serf and have the client nodes fetch the job via RPC.

From the [code](https://github.com/victorcoder/dkron/blob/f721d1c784c437d68cda286fe66a70d9ef60bd0e/dkron/rpc.go) I can tell that RPC is already used to communicate the results of an execution, so maybe transferring the job payload might be a good solution as well. 

What do you think about this idea? If you like it, we could contribute an implementation.

seiffert:
Fixes #122 

This is still work in progress.

## Todo

- [x] Include RPC server address in serf `QueryRunJob`
- [ ] Query `Execution` via RPC
- [ ] Remove `Execution` from `ExecutionRequest`
- [ ] Respond to `QueryRunJob` with an `ExecutionResponse` that does not include the `Execution` (because this might violate the serf query size as well) 
- [ ] Send `ExecutionStarted` via RPC (instead of including the started execution in the `QueryRunJob` response)
- [ ] Remove job size validation (introduced in #114)
- [ ] Tests

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -56,6 +56,11 @@ type AgentCommand struct {
 	candidate  *leadership.Candidate
 }
 
+type ExecutionRequest struct {
+	Execution Execution `json:"execution"`
+	RPCServer string    `json:"rpc_server"`
+}
+
 func (a *AgentCommand) Help() string {
 	helpText := `
 Usage: dkron agent [options]
@@ -578,12 +583,13 @@ func (a *AgentCommand) eventLoop() {
 						"at":      query.LTime,
 					}).Debug("agent: Running job")
 
-					var ex Execution
-					if err := json.Unmarshal(query.Payload, &ex); err != nil {
+					var req ExecutionRequest
+					if err := json.Unmarshal(query.Payload, &req); err != nil {
 						log.WithFields(logrus.Fields{
 							"query": QueryRunJob,
 						}).Fatal("agent: Error unmarshaling job payload")
 					}
+					ex := req.Execution
 
 					log.WithFields(logrus.Fields{
 						"job": ex.JobName,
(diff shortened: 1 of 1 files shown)


seiffert:
@victorcoder One open question is how to communicate the start of an execution. In the current implementation, the agent fills in some information on the execution and sends it back in the serf response. I imagine that the same size limits apply to this response. Therefore, I suggest to communicate the start of an execution via RPC as well. In this RPC-Request, the updated Execution could be communicated and the server node could save it. 

The result would be a workflow like this:

dkron runjob workflow

Do you have a better idea or feedback in general?

victorcoder:
@seiffert I'll try to reduce the roundtrips to minimum. Given that the serf query default timeout (15s) should be enough, we can use the same roundtrip to give back a reduced version of the execution with the needed values.

So the workflow could be as follows:

image

Take into account that

1. RunJob query sends: JobID & RPC config for calling the server
2. RunJob query responds with just: Start time, status

victorcoder:
@seiffert About the code, don't remove the query rpc config, as it's needed for the `QueryExecutionDone` query, consider that, when it's time to respond the execution done, the original server address that originated the execution could have vanished, so we can not discard this roundtrip.

seiffert:
Aaah, ok, now I see why `QueryRPCConfig` is needed. 

I'm still wondering where to create the `Execution` though. I think we need to add a `GetExecution` method to the store in order to create the Execution when triggering `QueryRunJob` and retrieve it from the store in `GetExecution`. This would also offer the side-benefit of being able to expose single executions via the Dkron API. As identifying attribute for executions, I'd suggest to use the `Execution.Group` property (as we already do in our code outside of Dkron). What do you think?

victorcoder:
@seiffert yep, as you discovered you can not use `Group` to get a single execution, the executions `Group` concept is to indentify/group executions, think about when you start an execution it could target more than one node, but it's handled as different `Executions` but the same execution `Group` as it started in exactly that moment.

I added a WIP page in the wiki to explain some design decissions on Dkron that help to explain the rationale behind some design decissions. https://github.com/victorcoder/dkron/wiki/Internals#execution-results

victorcoder:
Hi @seiffert any progress on this? do you need something more from my side?

victorcoder:
@seiffert Ping

victorcoder:
Implemented in #165

Issue #79, opened 2016-02-24
One-off jobs


whizz:
Feature idea: I have a use case, where I need to execute something sometime later, but just once and then the job should be deleted or archived. I would specify a date and time, when I want the job to run, with optional time window and dkron would make sure, the job will be run on the specified servers or just one of them within the given window. If the job succeeds, it would be deleted. If not, it will be retried n times in the given window, with exponential backoff, maybe. If at the end of the window it still did not run successfully, it would be deleted also, possibly triggering an alert.

victorcoder:
I like the idea, but :-1:  to delete the job on completion, if you don't have notifications enabled you won't be able to see the execution result.

whizz:
Sure. By delete I meant basically "do not run again and move somewhere out of sight". But obviously it would be valuable to have access to it if I wanted to.

digitalcrab:
Hello @victorcoder 

In my current company, we decided to try you Dkron. And one of the required features for us is an ability to have "one-time" cron's, so I have time to work on this issue.

Do you have requirements or understanding from your point of view how it should be done? Or I can create something by myself and then open pull request? 

Thank you for you time

victorcoder:
@digitalcrab Any help is welcome!

> Do you have requirements or understanding from your point of view how it should be done? Or I can create something by myself and then open pull request?

I've been thinking in this issue for some time and I have a roadmap for it, let's follow the conversation by email. Ping me "victor at victorcastell.com"?

digitalcrab:
Check an email. Thank you

victorcoder:
@digitalcrab I did a PR with the feature, it actually took just a couple of hours

PR #137, opened 2016-06-02
One off jobs

victorcoder:
Implements #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,4 @@ script:
   - go vet ./...
   # - fgt golint ./...
   - glide install
-  - go test -v ./dkron
-
-env:
-  global:
-    - GO15VENDOREXPERIMENT=1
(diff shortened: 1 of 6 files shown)


junneyang:
@victorcoder 
does this mean: @at <datetime>, schedule at <datetime> only once?

if i want  schedule immediately and only once,
how to ? 
thanks!

victorcoder:
> does this mean: @at , schedule at only once?

yes

> if i want schedule immediately and only once, how to ?

Create a disabled job, then use the dashboard or the API to run the job

talexu:
> > does this mean: @at , schedule at only once?
> 
> yes
> 
> > if i want schedule immediately and only once, how to ?
> 
> Create a disabled job, then use the dashboard or the API to run the job

Is it possible to create a job which can run immediately?
The currently solution need to invoke the API twice (create and run), which is not so graceful.

Issue #78, opened 2016-02-24
Plugable job types


victorcoder:
This is an open discussion on how to implement different job types.

Some use cases won't need or don't benefit from shell execution and specially with lots of jobs.

whizz:
I could imagine use cases for HTTP jobs, which would just send a HTTP request of given type (POST/GET/...) to given URL with optional payload. Though this can be pretty easily be done with current shell jobs + curl. 

The question is, whether dkron is a system that has ambition to handle job volume on the order of hundreds per second or more, where the added overhead of executing curl would be of concern.

victorcoder:
I could imagine another one to run docker containers. But again I'm with you @whizz it's very easy to do it using the command line.

> The question is, whether dkron is a system that has ambition to handle job volume on the order of hundreds per second or more, where the added overhead of executing curl would be of concern.

Good question, I would say no, at this point it's not ready to handle this volume. The main bottleneck would be the storage backend.

seiffert:
We would definitely be interested in this feature! Right now, we are using dkron behind our own API (of an internal PaaS for our developers), and use the dkron jobs only to schedule executions of our own command line tool which runs Docker containers on our clusters. 
With the ability to implement our own job types, we could teach dkron to talk to our PaaS' API directly (I guess). :+1:

sercand:
What about adding gRPC job executors, service definition would be something like below
```
message Execution {
	// JobName to execute
    string job_name = 1;
    // Payload is data for the job, Dkron currently does not support this
    bytes payload = 2;
}
message ExecutionResult{
    // Partial output of the execution
    bytes output = 1;
}
service DkronExecutor {
    rpc Invoke(Execution) returns (ExecutionResult);
}
```
running this type of jobs won't be difficult than running shell commands
```
conn, err := grpc.Dial(job.GrpcURL) //new extra field for the Job
executor = pb.NewDkronExecutorClient(conn)
result, err := executor.Invoke(context.Background(), &pb.Execution{JobName:job.Name})
```

redben:
One way that works without additional features from dkron right now is wrapping the specific job type "executor" in a cli command. Works nice. No coupling.

caiovmv:
I have tried running jobs every 30 seconds, but found that the scheduler minimum time is 1 minute, this behavior is true, if so how can I change that in the source code ?

[...]

PR #178, opened 2016-08-19
WIP: run jobs with remote gRPC server

sercand:
As I described on this [comment](https://github.com/victorcoder/dkron/issues/78#issuecomment-240836423), this PR implements executing jobs via gRPC call.


victorcoder:
Thanks for the contribution! Could you elaborate in the use case for this?

sercand:
@victorcoder thanks for the quick response.
My motivation of this PR is I believe running jobs via bash script is powerful but also is limited. It requires dependent applications installed on the same host. If you're using docker you have to bundle them with the image. Also, it is not safe to execute another process at the same host. Safest way is running a task via the network. Http is another solution but it lacks type safety and standardization. Using gRPC provides type safety, all you need is the implement "dkron.v1.DkronExecutor" on a server. Also, you can use TLS to authenticate clients. GRPC now powers many distributed solutions like docker containerd/swarm-kit, etcd3, tensorflow, Google Bigtable.
To conclude, I believe implementing gRPC based job execution is imported for distributed dkron.

victorcoder:
I don't want to sound like a show stopper but this change implies opening the job types issue and aside that I already have plans for this, I don't like the way it's implemented, adding a new property `grpc` to de job is not something that I want. It should be something more generic implemented in a more flexible fashion.

In case of running a task using docker you should call your scheduler API to start a task or call docker directly from the job command, run the job and exit.

I don't like to have a (listening server where you need to know the URL in advance) waiting to be calles to run some action specially using docker, doesn't seem like a good practice.

I wanted to let you know this before you put more work into it

sercand:
I don't understand what do you mean by need to know the URL in advance. I am running Kubernetes cluster so the URL is like "email-sender.default". It's not that difficult, everything is a microservice inside the cluster. The difficult part is knowing locations of TLS cert files and whether it is using TLS or not.

victorcoder:
@sercand the URL is not the main problem I see, as you said probably not a problem, on the other hand I don't like this implementation, I'm not saying it doesn't work but it's something related to job types and this must be designed somehow more generic.

With this PR we'll get command jobs and `grpc` jobs but what about other types of jobs?

I need to think on the job types topic a bit more before accepting PRs for extending job types.

In the meantime, refer to my previous comment: In case of running a task using docker you should call your scheduler API to start a task or call docker directly from the job command, run the job and exit.

victorcoder:
Closing this at the moment, this will be addressed soon.

Issue #180, opened 2016-09-02
Success and error count and timestamp not recorded


whizz:
I upgraded to 0.9.0 and it seems the job statistics are no longer updated. If you create a new job, even when the job runs, the /v1/jobs/JOBNAME endpoint returns this:

...,
"success_count": 0,
"error_count": 0,
"last_success": "0001-01-01T00:00:00Z",
"last_error": "0001-01-01T00:00:00Z",
...

The executions endpoint returns a list of executions fine.

whizz:
After a bit of investigation, the trouble seems to be at https://github.com/victorcoder/dkron/blob/master/dkron/store.go#L72

I'll try to take a stab at fixing it.

victorcoder:
Fixed in #181

PR #181, opened 2016-09-08
Fixes issue with job stats not being updated (#180)

whizz:
I am not sure why the code in L71-L75 was there, but it definitely overwrote the stats and they never got updated.


victorcoder:
good catch @whizz, the code was there in case of modifying a job definition using the REST API so the input job doesn't overwrite the existing job stats.

LGTM

Issue #184, opened 2016-09-13
panic: runtime error: index out of range on starting job


SergioFernandezStratio:
Hi,

I've setup Dkron on three nodes, acting as masters and agents using zookeeper over Centos6. Everything wents ok, but when I schedule a test job, agents crashes with following error:

```
INFO[2016-09-13T14:09:35+01:00] agent: Starting job                           job=apuntador node=node2
panic: runtime error: index out of range

goroutine 43 [running]:
panic(0xbb9dc0, 0xc82000e0c0)
	/Users/victorcoder/src/github.com/mxcl/homebrew/Cellar/go/1.6.3/libexec/src/runtime/panic.go:481 +0x3e6
github.com/victorcoder/dkron/dkron.(*Store).GetExecutions(0xc8201a2960, 0xc8201a0330, 0x9, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/victorcoder/src/github.com/victorcoder/dkron/dkron/store.go:243 +0x588
github.com/victorcoder/dkron/dkron.(*Store).SetExecution(0xc8201a2960, 0xc820070480, 0x0, 0x0, 0x0, 0x0)
	/Users/victorcoder/src/github.com/victorcoder/dkron/dkron/store.go:329 +0x674
github.com/victorcoder/dkron/dkron.(*AgentCommand).setExecution(0xc820050420, 0xc8200f4370, 0xac, 0xac, 0xc8201a0300)
	/Users/victorcoder/src/github.com/victorcoder/dkron/dkron/agent.go:602 +0x180
github.com/victorcoder/dkron/dkron.(*AgentCommand).RunQuery(0xc820050420, 0xc820071c80)
	/Users/victorcoder/src/github.com/victorcoder/dkron/dkron/queries.go:92 +0x15e6
github.com/victorcoder/dkron/dkron.(*Job).Run(0xc820056100)
	/Users/victorcoder/src/github.com/victorcoder/dkron/dkron/job.go:98 +0x3c0
created by github.com/victorcoder/dkron/cron.(*Cron).run
	/Users/victorcoder/src/github.com/victorcoder/dkron/cron/cron.go:159 +0x58a
```

Job "apuntador" consists in:

```
#!/bin/bash
#
DATE_T=`date +%c`
echo "It's date "$DATE_T > /tmp/registry.txt
```
Job definition is:

```
{
	"name": "apuntador",
	"schedule": "@every 5s",
	"shell": false,
[...]

PR #185, opened 2016-09-14
#184 Hotfix: Performs nil check to mitigate a panic raised as described in #184

sayden:
#184 Panic is thrown on store.go:243. It is accessing to a store.Store member of the type Store which is nil.

This fix is doing the nil checking to return an error if client is not found.

I have also take advantage of this PR to fix a panic raised in tests. It is generally not recommended and much more idiomatic to use t.Fatal in a test.


victorcoder:
Could you split the PR in 2 one with the test fix and the other one where we can discuss the `Client` error? Thanks!

victorcoder:
Superseded by #193

Issue #182, opened 2016-09-12
Crash when deleting a job


SamuelHassine:
Hello,

I just installed the last version of dkron (0.9.0) with the etcd backend. I used the default configuration.

`INFO[2016-09-13T00:16:35+02:00] agent: Dkron agent starting                   node=oxid-master
INFO[2016-09-13T00:16:35+02:00] agent: joining: [] replay: true               node=oxid-master
INFO[2016-09-13T00:16:35+02:00] api: Running HTTP server                      address=:8080 node=oxid-master
INFO[2016-09-13T00:16:35+02:00] agent: Running for election                   node=oxid-master
INFO[2016-09-13T00:16:35+02:00] agent: Listen for events                      node=oxid-master
INFO[2016-09-13T00:16:35+02:00] agent: Cluster leadership lost                node=oxid-master
INFO[2016-09-13T00:16:35+02:00] agent: Cluster leadership acquired            node=oxid-master
`

I created the same job as in the unit tests:

POST /v1/jobs
`{"name": "test_job", "schedule": "@every 2s", "command": "date", "disabled": false}`

The job is working but when I want to delete it:
DELETE /v1/jobs/test_job

The Dkron agent crashed with the following stacktrace:

`panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0xa0 pc=0x4893b5]

goroutine 4262 [running]:
panic(0xbb9dc0, 0xc820010110)
    /Users/victorcoder/src/github.com/mxcl/homebrew/Cellar/go/1.6.3/libexec/src/runtime/panic.go:481 +0x3e6
github.com/victorcoder/dkron/dkron.(_AgentCommand).processFilteredNodes(0xc8201501e0, 0x0, 0x0, 0x0, 0x0, 0xc820010f80, 0x0, 0x0)
    /Users/victorcoder/src/github.com/victorcoder/dkron/dkron/agent.go:561 +0xa5
github.com/victorcoder/dkron/dkron.(_AgentCommand).RunQuery(0xc8201501e0, 0xc8202f5b00)
    /Users/victorcoder/src/github.com/victorcoder/dkron/dkron/queries.go:31 +0xd1
github.com/victorcoder/dkron/dkron.(_Job).Run(0xc820166300)
    /Users/victorcoder/src/github.com/victorcoder/dkron/dkron/job.go:98 +0x3c0
created by github.com/victorcoder/dkron/cron.(_Cron).run
[...]

PR #186, opened 2016-09-14
Prevent crash when deleting a job with high frequency schedule

richard-julien:
Try to solve https://github.com/victorcoder/dkron/issues/182.

### DIFF
--- a/dkron/queries.go
+++ b/dkron/queries.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"github.com/hashicorp/serf/serf"
+	"github.com/docker/libkv/store"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


dkron/queries.go
```diff
@@ -25,6 +25,13 @@ func (a *AgentCommand) RunQuery(ex *Execution) {
 
 	job, _ := a.store.GetJob(ex.JobName)
 
+	//Job can be removed and the QuerySchedulerRestart not yet received.
+	//In this case, the job will be nil. Simply cancel the current execution
+	if job == nil {
+		log.Debug("Job was removed, cancelling this excution, waiting for QuerySchedulerRestart")
```
victorcoder: I prefer to say "Job not found", but LGTM

dkron/queries.go
```diff
@@ -25,6 +25,13 @@ func (a *AgentCommand) RunQuery(ex *Execution) {
 
 	job, _ := a.store.GetJob(ex.JobName)
```
victorcoder: `err` should be checked here to see if it's `store.ErrKeyNotFound` then logging

richard-julien: Agree that's a better way to do it. I will change that.

richard-julien: PR updated.

victorcoder:
@richard-julien You must run `go fmt` prior to commit, thanks!

richard-julien:
Done. Sorry again ... devenv config updated to do that automatically.

Issue #167, opened 2016-07-30
JS error when no jobs are registered


kevana:
Small error, looks like the dashboard expects an array but the Jobs API endpoint returns null. Before I change anything, which is more idiomatic Go: null or an empty array? 

<img width="366" alt="screen shot 2016-07-29 at 11 51 45 pm" src="https://cloud.githubusercontent.com/assets/981551/17268166/845b703a-55e7-11e6-8cd6-3dd04adea44d.png">

victorcoder:
Thanks for spotting this @kevana, well it's not Go related, it's related to RESTful API design. We have two options:

1. Return a 200 and an empty array
2. Return a 204 NO CONTENT and an empty body

I'll go for 1 at this point, being 2 more appealing to me though, It'll depend on how angular is able to handle a 204.

Let's go for 1.

richard-julien:
+1 for solution one

victorcoder:
Fixed in https://github.com/victorcoder/dkron/issues/217

PR #187, opened 2016-09-15
Initialize result array with 0 lenght

richard-julien:
Fixing https://github.com/victorcoder/dkron/issues/167. 
I'm a new go player, dont know if its the correct way of doing it.

### DIFF
--- a/dkron/store_test.go
+++ b/dkron/store_test.go
@@ -1,10 +1,10 @@
 package dkron
 
 import (
+	s "github.com/docker/libkv/store"
+	"github.com/stretchr/testify/assert"
 	"testing"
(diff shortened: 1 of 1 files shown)


dkron/store.go
```diff
@@ -186,7 +186,7 @@ func (s *Store) GetJobs() ([]*Job, error) {
 		return nil, err
 	}
 
-	var jobs []*Job
+	var jobs = make([]*Job, 0)
```
richard-julien: Yes of course, i will do that.

richard-julien: Test case provided. I hope it'll be ok

PR #188, opened 2016-09-15
If the command fail to start, error should be reported for statistics

richard-julien:
When the command provided for the job cannot start, for example if you used something missing in the node system (like trying to execute a curl command on windows), the job execution stats should be updated to reflect the execution failure to inform the administrator that something is broken.


victorcoder:
Please provide a little of context and description.

richard-julien:
Description updated.

Issue #210, opened 2016-11-24
Use DELETE or POST http verb for /leave API endpoint


kevana:
`GET` requests shouldn't cause side effects such as modifying cluster membership. I suggest using `DELETE` or `POST` to make clear that the leave endpoint is not a [safe operation](https://tools.ietf.org/html/rfc7231#section-4.2.1). I believe this change would require a version bump for the API.

[Source to change](https://github.com/victorcoder/dkron/blob/master/dkron/api.go#L55)

Thoughts?

victorcoder:
Good catch @kevana, it should be changed. As we're in pre final release here, the best option I see is to maintain the old method giving a deprecation message as WARNING in log output and add the new verb.

PR #214, opened 2016-12-04
Allow usage of POST for /leave

victorcoder:
Adding a deprecation warning message.

Use http package constants for http verbs


victorcoder:
fixes #210

PR #216, opened 2016-12-09
FIX docker-compose to use Zookeeper as coordinator

Sdedelbrock:
Docker-compose script does not run by default, the proposed changes would run Dkron with zookeeper as the default coordinator, and will not unnecessarily download consul/etcd

- commented out consul init script, uncommented zookeeper init script
- commented out etcd and consul services as they are not needed.
- added `Uncomment to use consul` comment above consul service
- added `Uncomment to use zk` comment above zookeeper service

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,22 +2,24 @@
 version: '2'
 services:
   # Uncomment to use etcd
-  etcd:
-    image: microbox/etcd
-    ports:
-      - "4001"
-    volumes:
-      - ./etcd.data:/data
-    command: -name=dkron1
+  #etcd:
+  #  image: microbox/etcd
+  #  ports:
+  #    - "4001"
+  #  volumes:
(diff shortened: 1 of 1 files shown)


victorcoder:
Thanks @Sdedelbrock but the compose file in dkron is intended for development/testing purposes, I'll usually develop and test switching between all the engines and specially using `etcd` for testing and `consul` for development so I'm not a `zookeeper` fan sorry.

Sdedelbrock:
Agreed, I am more of a consul fan myself.  Docker-compose is how I first evaluated Dkron, as recommended per the Readme.  It's a great way to evaluate the project.  However the docker-compose file on master is not functional.  The file has service-coordinators all mixed up (starting ETCD but using Zookeeper).  This pull request fixes that issue by running Zookeeper as the default.  It doesn't have to be that way, I had just chosen it as it seemed to be the last coordinator in use when the file was last edited.

PR #217, opened 2016-12-09
Return empty list instead of null for empty job listing

oldmantaiter:
- If we use a new namespace and then call `http://host:port/v1/jobs`, `null`
  will be returned instead of valid JSON `[]`


oldmantaiter:
I think this solves #167 - I noticed it when writing pydkron

Issue #112, opened 2016-04-29
Webhook alert firing on success


patrick-bark:
Hey, I just got the web hooks going, and they are working, except they are also firing when a job succeeds.

```
dkron-server-central | time="2016-04-29T22:22:12Z" level=info msg="agent: Running for election" node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:22:12Z" level=info msg="agent: Cluster leadership lost" node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:22:12Z" level=debug msg="agent: Stopping scheduler due to lost leadership" node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:30:03Z" level=debug msg="agent: Received event" event="query: rpc:config" node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:30:03Z" level=debug msg="agent: RPC Config requested" at=633 node=aws-nv-p-ops-elk payload= query="rpc:config" 
dkron-server-central | time="2016-04-29T22:30:06Z" level=debug msg="rpc: Received execution done" group=1461969000000382365 job=sched node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:30:06Z" level=debug msg="store: Retrieved job from datastore" job=sched node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:30:06Z" level=debug msg="store: Setting key" execution=1461969000000899257-aws-nv-p-ops-luigi.valkyrie.net job=sched node=aws-nv-p-ops-elk 
dkron-server-central | time="2016-04-29T22:30:06Z" level=debug msg="store: Setting job" job=sched json="{\"name\":\"sched\",\"schedule\":\"0 */5 * * *\",\"command\":\"/home/patrick.barker/.pyenv/versions/anaconda3-2.4.1/bin/python /home/patrick.barker/dkron-python/scriptname/sched.py\",\"owner\":\"\",\"owner_email\":\"patrick.barker@oracle.com\",\"run_as_user\":\"\",\"success_count\":1,\"error_count\":0,\"last_success\":\"2016-04-29T22:30:03.423389784Z\",\"last_error\":\"0001-01-01T00:00:00Z\",\"disabled\":false,\"tags\":{\"role\":\"luigi_docker\"}}" node=aws-nv-p-ops-elk 
[...]

PR #219, opened 2016-12-09
Add notification rules for jobs

oldmantaiter:
- Adds `notify_on` to the job defintion where either `success` or `failure` can
  be specified to supress output for that job depending on the execution outcome
	- If anything else other than `success` or `failure` is used, it will be
	  ignored
X-Github-Closes: #112

### DIFF
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -80,6 +80,9 @@ type Job struct {
 
 	// Processors to use for this job
 	Processors map[string]PluginConfig `json:"processors"`
+
+	// Notify condition (success/failure)
+	NotifyOn string `json:"notify_on"`
 }
 
 // Run the job
--- a/dkron/notifier.go
+++ b/dkron/notifier.go
@@ -31,6 +31,10 @@ func Notification(config *Config, execution *Execution, exGroup []*Execution, jo
 }
 
 func (n *Notifier) Send() {
+	if (n.Job.NotifyOn == "success" && !n.Execution.Success) ||
+		(n.Job.NotifyOn == "failure" && n.Execution.Success) {
+		return
+	}
 	if n.Config.MailHost != "" && n.Config.MailPort != 0 && n.Job.OwnerEmail != "" {
 		n.sendExecutionEmail()
 	}
--- a/dkron/notifier_test.go
+++ b/dkron/notifier_test.go
@@ -147,3 +147,87 @@ var templateTestCases = func(n *Notifier) []templateTestCase {
 		},
 	}
 }
+
+func TestNotifier_notifyOnSuccessPass(t *testing.T) {
+	notified := false
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		io.Copy(w, r.Body)
+		notified = true
+	}))
+	defer ts.Close()
+	c := &Config{
+		WebhookURL:     ts.URL,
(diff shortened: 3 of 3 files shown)


victorcoder:
This LGTM @oldmantaiter but as you may have noticed, the upcoming v0.9.2 comes with "processor" plugins, my idea was to move notifications to processor plugins, that would allow for greater flexibility.

That will avoid the necessity of modifying the `Job` struct to add a new property.

Tell me if this is something that you can live without until the email notification plugin is ready.

Thanks!

oldmantaiter:
@victorcoder Absolutely, I can definitely wait and then assist with plugins. I was thinking a specific slack one using the attachment message type vs having to add another template field like `{{.Color}}`

yebo29:
Is notify_on something we can use right now?

victorcoder:
Sadly one year after and I didn't had time to build the new processor. Help wanted

kimlood:
So, to understand correctly. 
In the current versions, the webhooks for slack for example is not working? Or how should this be configured? I have added all the config for webhooks according to the documentation, but still an execution of a job does not send an notification, not for any type (INFO/ERROR)

victorcoder:
@kimlood notifications in the current versions work. What's not implemented is notification rules `success/error`. 

Example config notifications for slack/email:

```
webhook-url: "https://hooks.slack.com/services/xxxxxxxxxxxxx"
webhook-payload: "payload={\"text\": \"{{.Report}}\", \"channel\": \"#notifications\"}"
webhook-headers: "Content-Type:application/x-www-form-urlencoded"
mail-host: "your.smpt.server.com"
mail-port: 25
mail-username: xxxx
mail-password: xxxx
mail-from: cron@yourdomain.com
mail-subject-prefix: "[Dkron-production]"
```

kimlood:
> @kimlood notifications in the current versions work. What's not implemented is notification rules `success/error`.
> 
> Example config notifications for slack/email:
> 
> ```
> webhook-url: "https://hooks.slack.com/services/xxxxxxxxxxxxx"
> webhook-payload: "payload={\"text\": \"{{.Report}}\", \"channel\": \"#notifications\"}"
> webhook-headers: "Content-Type:application/x-www-form-urlencoded"
> mail-host: "your.smpt.server.com"
> mail-port: 25
> mail-username: xxxx
> mail-password: xxxx
> mail-from: cron@yourdomain.com
> mail-subject-prefix: "[Dkron-production]"
> ```

Great, thanks @victorcoder!
And there are no other config that is needed for a job to send notifications? As long as this configuration is correct?

victorcoder:
@kimlood in case of email notifications, the job, must include the `owner_email` attribute with a valid email.

victorcoder:
This is now implemented in the Pro Email processor https://dkron.io/pro/email/.

I don't want to add new fields to the job structure for this and I want to keep core email notifications ultra simple, and probably removing it from the core at some point.

I know this decision could make some people angry but I can say that I will be opensourcing pro plugins at some point and this is on top.

yebo29:
email on success/failure vs just on every event is a fundamental feature to have IMO. I look forward to the open sourcing of the plugins. Any idea on the time frame? 
If I knew golang I would help, but my current situation does not allow me to do so (I'm full head-first with Python at the moment, along with Arduino and other projects. Yes, for work and pleasure. I'm still only human as much as I'd like to be Superman).

PR #226, opened 2017-01-09
A new parameter for advertise_rpc_port

firstway:
We had a issue when we run the dkron services in marathon framework.
Because the marathon/docker  have the mapping for the RPC server from container IP:Port to Host IP:Port, such as `10.0.2.3:6868 -> 172.217.1.78:13123` in which the port `13123` is dynamically assigned by docker, other agents can NOT access the 10.0.2.3:6868 since it's inside the container, the agents should connect  the port `:13123` instead of `:6868`.
We already have the -advertise address( IP+Port) for `8946`, but not for RPC port.
This PR added a new parameter `advertise_rpc_port` whose value should be '13123' in above example.
I am not 100% sure it's the best solution for the issue, but it does work in our marathon cluster.

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -89,6 +89,7 @@ Options:
   -ui-dir                         Directory from where to serve Web UI
   -rpc-port=6868                  RPC Port used to communicate with clients. Only used when server.
                                   The RPC IP Address will be the same as the bind address.
+  -advertise-rpc-port             Use the value of -rpc-port by default
 
   -mail-host                      Mail server host address to use for notifications.
   -mail-port                      Mail server port.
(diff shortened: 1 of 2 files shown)


victorcoder:
@firstway Have you tried just setting `rpc-port` to the desired port?

firstway:
@victorcoder  do you use containers in "host" mode? or "bridge" mode?
we use it in "bridge" mode. 
we could set a port for `rpc-port`, e.g. 7788. but in "bridge" mode, when we start a container (in marathon), it will give you a port mapping(like NAT)   `10.0.2.3:7788 -> 172.217.1.78:?????`, the dkron RPC server will listen on `:7788` inside the container (that's the -rpc-port is supposed to do), other agents only can access port "?????" , but the port '????' is not predictable. 
ref : https://www.ctl.io/developers/blog/post/docker-networking-rules/   : "Publishing ports with -P and EXPOSE"

victorcoder:
Right @firstway I use it in "bridge" mode. Then my question is, how do you know what "advertise-rpc-port" will be assigned and therefore, what you need to set beforehand in config?

I need to think a bit more on this topic

firstway:
@victorcoder , yes, we don't know the ports assigned by marathon before we start the containers(because we don't want to use fixed ports).but we can use some `Environment Variables` for these ports in the config file( e.g.  -advertise-rpc-port  $PORT2).
https://mesosphere.github.io/marathon/docs/ports.html  #Environment Variables

victorcoder:
Thanks @firstway!

PR #234, opened 2017-01-27
Leverage docker layer cache to build dkron image

Eyjafjallajokull:
This patch improves docker image build time, by leveraging docker layer
cache. Dependencies installation step can be skipped if files glide.lock
and glide.yaml were not changed since last build. Cheers


victorcoder:
Thx!

Issue #247, opened 2017-03-25
Orphan plugin processes


m1r4ge:
After restarting dkron process the plugin processes continue to execute and dkron creates duplicate ones (dkron-processor-files, dkron-processor-log, dkron-processor-syslog)

victorcoder:
Please, can you report your OS and dkron version?

m1r4ge:
OS: Ubuntu 16.04.2 LTS (Xenial Xerus), Dkron v0.9.3

victorcoder:
I'll do some checks, thanks for reporting

m1r4ge:
I restarted dkron with Ctrl+C ;)

stephenranjit:
facing the same issue. as a workaround, iIm killing long-running processes using a script but it would be nice if processor processes exit along with dkron.

OS: RedHat 6 2.6.32-642.6.1.el6.x86_64
Dkron v0.9.3

```
<user> 538     1  0 May24 ?        00:00:00 dkron-processor-files
<user> 548     1  0 May26 ?        00:00:00 dkron-processor-files
<user> 687     1  0 May24 ?        00:00:00 dkron-processor-log
<user> 704     1  0 May24 ?        00:00:00 dkron-processor-syslog
```

pabdavis:
Seeing same issue occurs after crash of dkron as well as Ctrl + C of running process

OS: Centos 7 
Dkron v0.9.3

victorcoder:
FTR https://github.com/hashicorp/go-plugin/pull/81

PR #277, opened 2017-06-25
clean up plugins upon recieving SIGINT and SIGTERM

danielhan:
This PR addresses https://github.com/victorcoder/dkron/issues/247.  I am seeing the same results in RHEL and OSX.  CleanupClients() was not being called after a SIGINT or SIGTERM.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,7 +38,6 @@ func main() {
 	plugins.DiscoverPlugins()
 
 	// Make sure we clean up any managed plugins at the end of this
-	defer plugin.CleanupClients()
 
 	c.Commands = map[string]cli.CommandFactory{
 		"agent": func() (cli.Command, error) {
@@ -64,8 +63,8 @@ func main() {
 	exitStatus, err := c.Run()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error executing CLI: %s\n", err.Error())
-		os.Exit(1)
 	}
 
+	plugin.CleanupClients()
 	os.Exit(exitStatus)
 }


main.go
```diff
@@ -61,11 +60,25 @@ func main() {
...
+
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		<-sigs
+		done <- 0
+	}()
+
+	go func() {
+		exitStatus, err := c.Run()
```
danielhan: Putting this in a goroutine may seem a bit odd.  Initially, I tried keeping this as part of the implicit/main goroutine, only spinning off another goroutine for the signal handler.  In the signal handler goroutine, I was calling plugin.CleanupClients(), then doing an os.Exit().  I ran into rather odd behavior, where the plugin.CleanupClients() would invoke, but the system would exit before plugin.CleanupClients() finished.

victorcoder:
Thanks for the PR @danielhan I did some tests on this and it's failing as you and others reported.

The problem is that the `defer` function is never executed when calling `os.Exit`, I tested a quick fix just moving the `plugin.CleanupClients()` just before the last `os.Exit(exitStatus)`

Signals are handled in https://github.com/victorcoder/dkron/blob/master/dkron/agent.go#L325 and I find more complicated what you're proposing.

What do you think in adapting this PR with this solution?

danielhan:
@victorcoder I missed the signal handling code!  Sorry about that.  I think your idea for the quick fix essentially does the same thing I had written, without having to spin off another goroutine.  It is much cleaner!

main.go
```diff
@@ -64,8 +63,8 @@ func main() {
 	exitStatus, err := c.Run()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error executing CLI: %s\n", err.Error())
-		os.Exit(1)
```
danielhan: Note that previously, the exit code of AgentCommand.Run() was not being returned after an error.  The returned exit code was either 0 or 1.  My current changes have the system returning the exit code generated by AgentCommand.Run().  I'm not sure if this is the behavior you want

victorcoder:
Thanks @danielhan

PR #298, opened 2017-09-06
fixed GetLastExecutionGroup

firstway:
Store.GetLastExecutionGroup() could return a random execution because the `executions list` is not in order. This PR fixed this issue by comparing ~~sorting~~ them by `StartedAt` time.

### DIFF
--- a/dkron/store.go
+++ b/dkron/store.go
@@ -274,16 +274,23 @@ func (s *Store) GetLastExecutionGroup(jobName string) ([]*Execution, error) {
 	if err != nil {
 		return nil, err
 	}
-	if len(res) == 0{
+	if len(res) == 0 {
 		return []*Execution{}, nil
 	}
 
+	var lastEx Execution
 	var ex Execution
-	err = json.Unmarshal([]byte(res[len(res)-1].Value), &ex)
-	if err != nil {
-		return nil, err
+	// res does not guarantee any order, so sort them by `StartedAt` time
+	for _, node := range res {
+		err := json.Unmarshal([]byte(node.Value), &ex)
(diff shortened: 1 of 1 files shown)


victorcoder:
Hi @firstway Thanks for the work, I find this implementation to be not optimized for performance. What do you think in implementing it using the https://golang.org/pkg/sort/ on the `res` ?

firstway:
Hi @victorcoder , actually , I don't sort execution array, I just compare them by `StartedAt` time  with O(n) time. I changed the comment which might be a little bit misunderstanding.

dkron/store.go
```diff
@@ -274,16 +274,24 @@ func (s *Store) GetLastExecutionGroup(jobName string) ([]*Execution, error) {
...
 		return nil, err
 	}
-	if len(res) == 0{
+	if len(res) == 0 {
 		return []*Execution{}, nil
 	}
 
+	var lastEx Execution
 	var ex Execution
-	err = json.Unmarshal([]byte(res[len(res)-1].Value), &ex)
```
firstway: this original line assumes that the LAST item of `res` is the LATEST Execution, the fact is that `res` is **unsorted**, so you have to iterate all Executions to find the LATEST one.

firstway:
Can you merge this, @victorcoder 
because `Job` needs `GetLastExecutionGroup` to get the current status https://github.com/victorcoder/dkron/blob/master/dkron/job.go#L125-L131, without this fix, the status could be wrong and dashboard will show incorrect info for the running jobs.

victorcoder:
Thanks @firstway

PR #317, opened 2017-11-28
Update plugins.go

ti:
change  bitbucket.org/kardianos/osext to github.com/kardianos/osext, the new bitbucket.org/kardianos/osext not supported


victorcoder:
Please, update https://github.com/victorcoder/dkron/blob/master/Gopkg.toml#L24 too.

Thanks

Issue #279, opened 2017-06-29
Job dependencies reset when updating parent job.


tunasalat:
When I update existing job that has dependent jobs, the property dependent_jobs becomes null and child jobs stop executing. I use consul as datastore and dkron 0.9.4. See details below.

```
# cat test_job.js 
{
    "name": "job_test",
    "command": "sleep 1",
    "schedule": "0 0/2 * * * *",
    "tags": {
            "host": "host1"
    },
    "retries" : 2,
    "concurrency": "forbid"
}
# cat test_job2.js 
{
    "name": "job_test_chain",
    "parent_job": "job_test",
    "command": "sleep 2",
    "tags": {
            "host": "host1"
    },
"retries" : 2,
"concurrency": "forbid"
}
# curl -v -n -X GET $HOST/v1/jobs  | python -m json.tool

[]


# curl -v -n -X POST $HOST/v1/jobs  -H "Content-Type: application/json" --data-binary "@test_job.js" 
{"name":"job_test","schedule":"0 0/2 * * * *","shell":false,"command":"sleep 1","owner":"","owner_email":"","success_count":0,"error_count":0,"last_success":"0001-01-01T00:00:00Z","last_error":"0001-01-01T00:00:00Z","disabled":false,"tags":{"host":"host1"},"retries":2,"dependent_jobs":null,"parent_job":"","processors":null,"concurrency":"forbid"}


# curl -v -n -X POST $HOST/v1/jobs  -H "Content-Type: application/json" --data-binary "@test_job2.js"
{"name":"job_test_chain","schedule":"","shell":false,"command":"sleep 2","owner":"","owner_email":"","success_count":0,"error_count":0,"last_success":"0001-01-01T00:00:00Z","last_error":"0001-01-01T00:00:00Z","disabled":false,"tags":{"host":"host1"},"retries":2,"dependent_jobs":null,"parent_job":"job_test","processors":null,"concurrency":"forbid"}


# curl  -n -X GET $HOST/v1/jobs  | python -m json.tool
[
    {
        "command": "sleep 1",
        "concurrency": "forbid",
        "dependent_jobs": [
            "job_test_chain"
        ],
        "disabled": false,
        "error_count": 0,
        "last_error": "0001-01-01T00:00:00Z",
        "last_success": "2017-06-29T11:32:01.077349261+02:00",
        "name": "job_test",
        "owner": "",
        "owner_email": "",
        "parent_job": "",
        "processors": null,
        "retries": 2,
[...]

Issue #312, opened 2017-10-23
tag filtering is working like wildcard on tag's value


genepix:
Hello,

I'm facing a funny scenario here, a job with tag condition {"app": "myapp"} actually match nodes with {"app": "myapp"}, but also nodes with {"app": "myappxxxxxxxxxxxxx"}.

I don't know golang but nothing seems odd to me at the "if" statement in the processFilteredNodes function (where i guess it's the filter part) : https://github.com/victorcoder/dkron/blob/master/dkron/agent.go#L610

tested with 0.9.4 and 0.9.5 release

genepix:
Hello,
I've investigated a bit more about this issue : 

This function returns an empty list of nodes, so serf's query behaviour is based on tag's filter (due to empty nodes list). But Serf acts like a wildcard regexp for tag filtering : 

https://github.com/hashicorp/serf/blob/master/serf/query.go#L227

A dirty workaround is to set the wished wildcard in the job (like {"app":"^mystrictapp$"}), but the count filter won't work.

victorcoder:
Good catch @genepix, thanks for the investigation, I'll fix this

genepix:
De nada :) !

Issue #77, opened 2016-02-24
scheduling mismatch tags with same prefix


schemacs:
I have two servers, one has role of **vpnserver** and another **vpnserver2**, after posting a job targeting tags **vpnserver**, both servers got scheduled.

> curl 127.0.0.1:8080/v1/members | python -m json.tool
> [
>     {
>         "Addr": "10.0.0.9",
>         "DelegateCur": 4,
>         "DelegateMax": 4,
>         "DelegateMin": 2,
>         "Name": "h1",
>         "Port": 8946,
>         "ProtocolCur": 2,
>         "ProtocolMax": 2,
>         "ProtocolMin": 1,
>         "Status": 1,
>         "Tags": {
>             "key": "xxx1",
>             "role": "vpnserver2",
>             "server": "true"
>         }
>     },
>     {
>         "Addr": "10.0.0.1",
>         "DelegateCur": 4,
>         "DelegateMax": 4,
>         "DelegateMin": 2,
>         "Name": "h2",
>         "Port": 8946,
>         "ProtocolCur": 2,
>         "ProtocolMax": 2,
>         "ProtocolMin": 1,
>         "Status": 1,
>         "Tags": {
>             "key": "xxx2",
>             "role": "vpnserver",
>             "server": "true"
>         }
>     }
> ]
> 
> curl -n -X POST 127.0.0.1:8080/v1/jobs \
>   -H "Content-Type: application/json" \
>  \
>   -d '{
>   "name": "uptime",
>   "schedule": "0 30 \* \* \* *",
>   "command": "uptime",
>   "tags": {
>     "role": "vpnserver"
>   }
> }'

victorcoder:
@schemacs note that you can use curl `127.0.0.1:8080/v1/members?pretty`

Could you post an extract of the execution results? `127.0.0.1:8080/v1/executions/uptime?pretty`

schemacs:
Thanks, below is what I got after calling the api: 

> [
>         {
>                 "job_name": "uptime",
>                 "started_at": "2016-02-26T09:30:00.293483665+08:00",
>                 "finished_at": "2016-02-26T09:30:00.318768786+08:00",
>                 "success": true,
>                 "node_name": "h1",
>                 "group": 1456450200000444958,
>                 "job": {
>                         "name": "uptime",
>                         "schedule": "0 30 * * * *",
>                         "command": "uptime",
>                         "owner": "",
>                         "owner_email": "",
>                         "run_as_user": "",
[...]

PR #322, opened 2018-01-10
[FIX] copy dependent job from previous job to new job version

luhhujbb:
Fix issue #279 

A way to reproduce the bug was
1 °) POST parentjob.json
2°) POST childjob.json

Here dependentJob is successfully fullfilled

3°) POST parentjob.json (in order to update it)

dependentJobs have been wiped
 
Now setJob function also takes previousJob as arg (eventually niled) and if not niled copy dependent jobs (if present)

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -126,7 +126,7 @@ func (a *AgentCommand) jobCreateOrUpdateHandler(c *gin.Context) {
 	}
 
 	// Save the job to the store
-	if err = a.store.SetJob(&job); err != nil {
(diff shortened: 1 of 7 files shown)


luhhujbb:
The second fix is related to issue #77 and #312 , serf use regexp matching so the trick is to add ^ at the beginning of tags value and a $ at the end of tags value when creating serf query.

victorcoder:
Thanks for the work @luhhujbb, coud you rebase with master?

luhhujbb:
Hi, I've just rebase from master

Issue #237, opened 2017-02-01
Cluster missing executions - "dial tcp: missing address"


alexef:
# My Setup

5 dkron servers, 2 dkron runners. 

Using Consul as the data store.

# Problem

From time to time, randomly, I see in the runners log:

> time="2017-02-01T11:39:30+01:00" level=info msg="agent: Starting job" job="cjms_fetch" node=watcher11 
> time="2017-02-01T11:39:34+01:00" level=error msg="rpc: error dialing." err="dial tcp: missing address" node=watcher11 server_addr= 
> time="2017-02-01T11:39:34+01:00" level=error msg="agent: Error invoking job command" error="dial tcp: missing address" node=watcher11 
> time="2017-02-01T11:40:30+01:00" level=info msg="agent: Starting job" job="cjms_check" node=watcher11 

In the execution history, we just see the started and no output:

<img width="1169" alt="screen shot 2017-02-01 at 12 58 23" src="https://cloud.githubusercontent.com/assets/346576/22505963/353a9592-e87e-11e6-8df3-861c20229a30.png">

victorcoder:
Hi @alexef thanks for reporting, I haven't seen this error before, I'm guessing that some of the servers could be out of the serf cluster.

Could you paste here the output of `/v1/members` of the node?

alexef:
```[
	{
		"Name": "watcher10",
		"Addr": "x.x.x.x",
		"Port": 8946,
		"Tags": {
			"dkron_version": "0.9.3-b1",
			"environment": "production",
			"role": "watcher"
		},
		"Status": 1,
		"ProtocolMin": 1,
		"ProtocolMax": 2,
		"ProtocolCur": 2,
		"DelegateMin": 2,
		"DelegateMax": 4,
		"DelegateCur": 4
	},
	{
		"Name": "worker11",
		"Addr": "x.x.x.x",
		"Port": 8946,
		"Tags": {
			"dkron_version": "0.9.3-b1",
			"environment": "production",
			"role": "worker"
		},
		"Status": 1,
		"ProtocolMin": 1,
		"ProtocolMax": 2,
		"ProtocolCur": 2,
		"DelegateMin": 2,
		"DelegateMax": 4,
		"DelegateCur": 4
	},
	{
		"Name": "pillar14",
		"Addr": "x.x.x.x",
		"Port": 8946,
		"Tags": {
			"dkron_server": "true",
			"dkron_version": "0.9.3-b1",
			"environment": "production",
[...]

PR #336, opened 2018-03-08
Add retries for rpcconfig query timeouts

sysadmind:
This should fix an issue that I have had very similar
to #237 where tasks would fail to register their results
because they couldn't get the RPC address of a server. This
coupled with a job config of "concurrency": "forbid" meant
jobs would stop indefinitely.

The path where rpcAddr is nil from queryRPCConfig() is
a path dkron should never return because it means that
results are never recorded.

### DIFF
--- a/dkron/queries.go
+++ b/dkron/queries.go
@@ -150,6 +150,7 @@ func (a *AgentCommand) schedulerRestartQuery(leaderName string) {
 // Broadcast a query to get the RPC config of one dkron_server, any that could
 // attend later RPC calls.
 func (a *AgentCommand) queryRPCConfig() ([]byte, error) {
+queryConfigStart:
 	nodeName := a.selectServer().Name
 
 	params := &serf.QueryParam{
@@ -196,6 +197,13 @@ func (a *AgentCommand) queryRPCConfig() ([]byte, error) {
 	log.WithFields(logrus.Fields{
 		"query": QueryRPCConfig,
 	}).Debug("proc: Done receiving acks and responses")
+
+	if rpcAddr == nil {
+		log.WithFields(logrus.Fields{
+			"query": QueryRPCConfig,
+		}).Debug("proc: Failed to receive a config response, retrying")
+		goto queryConfigStart
+	}
 
 	return rpcAddr, nil
 }


dkron/queries.go
```diff
@@ -197,5 +198,12 @@ func (a *AgentCommand) queryRPCConfig() ([]byte, error) {
 		"query": QueryRPCConfig,
 	}).Debug("proc: Done receiving acks and responses")
 
+	if rpcAddr == nil {
+		log.WithFields(logrus.Fields{
+			"query": QueryRPCConfig,
+		}).Debug("proc: Failed to receive a config response, retrying")
+		goto queryConfigStart
```
victorcoder: This is a pretty hardcore looping, I prefer to add a retry limit control here then fataling, otherwise agents could end up in an endless loop trying to contact inexistent servers.

giaplv57: Hi,
Just wonder do we have any fix on this error?
I got into this error ("dial tcp: missing address") quite frequent: around 1 time/1hour in average.
Cannot apply dkron to production with this error :( ?

sysadmind: I agree that this could end up in an infinite loop. I'm concerned of the case of the `concurrency: forbid`job though. If there is a limited number of retries, and we stop trying to record that a job finished and the job will never run until the previous run is complete, we just stop executing a job. Maybe there are some alternatives that could be implemented like a timeout parameter on a job? This could give the best of both worlds.

victorcoder: The problem is not that the job doesn't finish, it does, but it is unable to communicate with a server. Under my production environment, this doesn't happen, so there should be some subtle network related issues that prevent this query to receive a response in time.

A timeout for waiting for a job execution is one possibility, but you will never know if the job never finished or it finished but you didn't get the response.

We can try several things here, but I will add a timeout here to unblock the PR.

victorcoder: @sysadmind @giaplv57 I'm unable reproduce the error, could you try to increase here https://github.com/victorcoder/dkron/blob/master/dkron/agent.go#L222 adding:

```
serfConfig.QueryTimeoutMult = (something > 16)
```

And see if it solves the problem?

Send me an email if you can't build dkron yourself.

Thanks

sysadmind: I tried this and set the timeout to 30. It didn't make a difference and within a few minutes I had jobs that failed to record their results with log messages about failing to dial a null address. 

I have a theory that it involves the master dropping the query, but I haven't had time to day to troubleshoot. I will see if I can create more steps to reproduce the exact issue.

victorcoder:
See #354

PR #339, opened 2018-03-28
Slightly improve processing of last execution group

sysadmind:
This helps to speed up the UI some as the /dashboard/jobs
endpoint needs to find the last execution group of every
job to display some UI components. This removes the call
to s.GetExecuionGroup becuase we already have that information
so there is no need to recursively get all execution groups
from the KV Store again and increase time and resource
utilization.


dkron/store.go
```diff
@@ -289,9 +290,12 @@ func (s *Store) GetLastExecutionGroup(jobName string) ([]*Execution, error) {
 		}
 		if ex.StartedAt.After(lastEx.StartedAt) {
 			lastEx = ex
+			executions = []*Execution{&ex}
+		} else if ex.Group == lastEx.Group {
```
victorcoder: I don't think this is going to work, given the first execution found is from a non last group it will end up in a Execution slice mixed with other executions found afterwards.

Could you provide a test?

sysadmind: I have added tests that I believe cover the potential cases. I did find an issue with a reference being changed in the loop, so I moved the variable initialization inside the loop like `GetExecutionGroup()` does. I also updated the test file slightly to make it easier to reduce repetition.

sysadmind:
How does this look for testify?

PR #347, opened 2018-04-12
Remove api understanding of job dependencies

sysadmind:
This should keep the api from needing to understand
that a job can have a parent or dependent jobs. This
slightly simplifies the functions and allows SetJob on
the store to have controll over job editing.

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -142,27 +142,8 @@ func (h *HTTPTransport) jobCreateOrUpdateHandler(c *gin.Context) {
 	}
 	c.BindJSON(&job)
 
-	// Get if the requested job already exist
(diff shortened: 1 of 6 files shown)


sysadmind:
This also removes the api call to getjob which gets repeated in setjob, so this should reduce the overall overhead of saving a job through the api.

sysadmind:
Is this failing test something that happens occasionally? It seems odd that my changes would have effected this.

victorcoder:
@sysadmind kind of liking it, could you rebase with `v0.10.0`

sysadmind:
Rebased and with the atomicjobput addition, these tests now pass and there isn't a deadlock.

PR #357, opened 2018-05-02
Add support for passing payload to command STDIN

gustavosbarreto:


victorcoder:
Thanks @gustavosbarreto, could you rebase with v0.10.0?

gustavosbarreto:
@victorcoder I'm not sure how to do it on v0.10.0.
The ```payload``` should be placed into ```Job``` struct or into specific ```executor_config```?

victorcoder:
@gustavosbarreto good question, as you can see, shell options in `Job` structure is being deprecated in favor of the `shell-executor` plugin so, we should add the `Payload` handling in the plugin not in the `Job` structure.

otavio:
@victorcoder, @gustavosbarreto and I were discussing it and we'll follow your advice to add it inside the plugin however we noticed that the protobuf generated code seems outdated in 0.10.0 branch.

Could you take a look?

victorcoder:
@otavio What do you mean outdated? you can try to run `protoc -I proto/ proto/executor.proto --go_out=plugins=grpc:dkron/` no changes on my side, can you check?

victorcoder:
@otavio are you still on this? should I take over?

otavio:
Yes please take it over; we ended not using it in our service as we replaced it with machinary.

PR #358, opened 2018-05-02
Add http executor

gustavosbarreto:
executor_config options:

url: url of http request
type: request method (any of https://golang.org/pkg/net/http/#pkg-constants)
body: request body encoded in base64
content_type: request content type

### DIFF
--- a/builtin/bins/dkron-executor-http/http.go
+++ b/builtin/bins/dkron-executor-http/http.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"bytes"
+	"encoding/base64"
+	"errors"
+	"io/ioutil"
+	"log"
+	"net/http"
+
+	"github.com/armon/circbuf"
+	"github.com/victorcoder/dkron/dkron"
+)
+
+const (
+	// maxBufSize limits how much data we collect from a handler.
+	// This is to prevent Serf's memory from growing to an enormous
+	// amount due to a faulty handler.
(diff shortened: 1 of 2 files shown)


victorcoder:
@gustavosbarreto would be nice to have this executor for the next release. Do you think you can take care of the polishing?

gustavosbarreto:
@victorcoder What do I need to improve?

builtin/bins/dkron-executor-http/http.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type Http struct {
+}
+
+// Process method of the plugin
+func (s *Http) Execute(args *dkron.ExecuteRequest) ([]byte, error) {
+	output, _ := circbuf.NewBuffer(maxBufSize)
+
+	if args.Config["url"] == "" {
```
victorcoder: What will happen if url is not set?

builtin/bins/dkron-executor-http/http.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil, errors.New("url is empty")
+	}
+
+	if args.Config["method"] == "" {
```
victorcoder: Same here

ProPheT777:
do we have a chance to see it in next release ? :pray: :heart:

redben:
Is it possible to add headers?

Issue #366, opened 2018-05-16
API returns error 400 when creating new job (v0.10.0)


gromo:
Hi, server returns error when creating tasks - status 400, although task is created.

Console output:
```
dkron_1   | time="2018-05-16T10:24:23Z" level=debug msg="store: Setting job" job=job-01 json="{\"name\":\"job-01\",\"timezone\":\"\",\"schedule\":\"* 10 * * * *\",\"shell\":false,\"command\":\"ls /tmp\",\"environment_variables\":null,\"owner\":\"\",\"owner_email\":\"\",\"success_count\":0,\"error_count\":0,\"last_success\":\"0001-01-01T00:00:00Z\",\"last_error\":\"0001-01-01T00:00:00Z\",\"disabled\":false,\"tags\":null,\"retries\":0,\"dependent_jobs\":null,\"parent_job\":\"\",\"processors\":null,\"concurrency\":\"allow\",\"executor\":\"\",\"executor_config\":null,\"status\":\"\"}" node=25ab53ffe49a
dkron_1   | [GIN-debug] [WARNING] Headers were already written. Wanted to override status code 400 with 201
```

Data to be sent:
```
{"name":"job-01","schedule":"* 10 * * * *","command":"ls \/tmp","concurrency":"allow","dependent_jobs":null,"disabled":false,"error_count":"0","last_error":null,"last_success":null,"owner":"","owner_email":"","parent_job":"","processors":null,"retries":0,"shell":false,"success_count":0,"tags":null}
```

gromo:
Hi @victorcoder , thanks for you note.

But problem is still there and not related to read-only fields. Another case when API tries to overwrite headers in v0.10.0:

Json to be sent:
```json
{"name":"simple","schedule":"* 5 * * * *","concurrency":"allow","dependent_jobs":null,"disabled":false,"executor":"shell","executor_config":{"command":"echo 1","shell":true},"owner":"","owner_email":"","parent_job":"","processors":{"log":{"forward":true}},"retries":0,"tags":null,"timezone":""}
```
Dkron console output:
```
[...]

PR #367, opened 2018-05-17
stop Job update execution on JSON parse error

gromo:
fix issue https://github.com/victorcoder/dkron/issues/366


dkron/api.go
```diff
@@ -138,7 +138,13 @@ func (h *HTTPTransport) jobCreateOrUpdateHandler(c *gin.Context) {
 	job := Job{
 		Concurrency: ConcurrencyAllow,
 	}
-	c.BindJSON(&job)
+
+	// Parse values from JSON
+	if err := c.BindJSON(&job); err != nil {
+		c.Writer.WriteString(err.Error())
```
digitalcrab: I would not expose this error, for the outside world

PR #378, opened 2018-05-23
Remove unnecessary updates of job finish times

sysadmind:
agent.RefreshJobStatus() was added to prevent missed exeuctionDone
messages from blocking forbidden concurrency jobs. This checked all
executions to see if they were done before responding to the job status.
If the jobs were already recorded as finished, this is pointless and
causes a lot of extra overhead on the network for no gain. The
extra effort also caused the execution FinishedAt parameter to
be set to a value that is not accurate.

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -574,7 +574,15 @@ func (a *Agent) RefreshJobStatus(jobName string) {
 	execs, _ := a.Store.GetLastExecutionGroup(jobName)
 	nodes := []string{}
 
+	unfinishedExecutions := []*Execution{}
 	for _, ex := range execs {
+		if ex.FinishedAt.IsZero() {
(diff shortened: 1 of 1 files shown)


dkron/agent.go
```diff
@@ -574,7 +574,16 @@ func (a *Agent) RefreshJobStatus(jobName string) {
 	execs, _ := a.Store.GetLastExecutionGroup(jobName)
 	nodes := []string{}
 
+	unfinishedExecutions := []*Execution{}
 	for _, ex := range execs {
+		if ex.FinishedAt.After(ex.StartedAt) {
```
victorcoder: This could be simplified as:
```
if ex.FinishedAt.IsZero() {
  unfinishedExecutions = append(unfinishedExecutions, ex)
}
```

sysadmind:
Good point on the optimization. New commit pushed.

PR #381, opened 2018-05-23
Mitigate bug of job stuck running

sysadmind:
I'm not sure what caused this bug for me, but from the API, I
could query /v1/jobs and all statuses were "success" but when
the scheduler would try and run the job, it would log that it was
skipping execution because concurrency was forbidden and the job
was running. This updates what status is used for a job and instead of
using the status on the pointer, querries the status instead to have
a real time view of the job.

### DIFF
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -251,18 +251,16 @@ func (j *Job) isRunnable() bool {
 	if j.Concurrency == ConcurrencyForbid {
 		j.Agent.RefreshJobStatus(j.Name)
 	}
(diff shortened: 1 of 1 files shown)


victorcoder:
@sysadmind could you chek https://github.com/victorcoder/dkron/pull/383 ? We're fixing the same issue.

sysadmind:
That should work. It's the same change I made. I don't know that I can replicate it because I don't know how it got stuck in the running state.

sysadmind:
Closing in favor of #383

PR #382, opened 2018-05-23
Add last run to cron timing logic

sysadmind:
Previously, if dkron was down during the window that a job should run
and then started back up, the job would be scheduled at the next
interval or schedule from that point as if the job had just run.

This changes that functionality to schedule jobs based on their
last run time instead of the current time so that on startup,
missed jobs will run.

### DIFF
--- a/cron/cron.go
+++ b/cron/cron.go
@@ -83,23 +83,23 @@ type FuncJob func()
 func (f FuncJob) Run() { f() }
 
 // AddFunc adds a func to the Cron to be run on the given schedule.
-func (c *Cron) AddFunc(spec string, cmd func()) error {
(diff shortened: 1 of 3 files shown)


victorcoder:
@sysadmind be careful, this is a very important change, missed runs are quite tricky to handle. It will need to give the user the option to choose if they want it to happen or not. 

Check the `epsilon` param in https://mesos.github.io/chronos/docs/api.html#adding-a-scheduled-job

I prefer not to introduce this change right now, unless a lot of care is dedicated to it.

victorcoder:
Closing because it needs adaptation to the new cron package version. Feel free to redo it from here.

PR #408, opened 2018-07-25
add filter ability on jobs api

kevynhale:
Looking for a review on the concept, still missing update to DOCS and tests.

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -122,7 +122,22 @@ func (h *HTTPTransport) indexHandler(c *gin.Context) {
 }
 
 func (h *HTTPTransport) jobsHandler(c *gin.Context) {
-	jobs, err := h.agent.Store.GetJobs(&JobOptions{ComputeStatus: true})
+	// Init the Job object with defaults
+	jobOptions := JobOptions{
+		ComputeStatus: true,
+	}
+
+	// Only parse values if there is a body.
+	// If there is not a body, retrieve all jobs.
+	if c.Request.ContentLength > 0 {
+		if err := c.BindJSON(&jobOptions); err != nil {
+			c.Writer.WriteString("Incorrect or unexpected parameters")
+			log.Error(err)
+			return
+		}
+	}
+
+	jobs, err := h.agent.Store.GetJobs(&jobOptions)
 	if err != nil {
 		log.WithError(err).Error("api: Unable to get jobs, store not reachable.")
 		return
--- a/dkron/store.go
+++ b/dkron/store.go
@@ -45,6 +45,7 @@ type Store struct {
 
 type JobOptions struct {
 	ComputeStatus bool
+	Tags          map[string]string `json:"tags"`
 }
 
 func init() {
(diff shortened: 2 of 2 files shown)


dkron/store.go
```diff
@@ -242,6 +244,26 @@ func (s *Store) GetJobs(options *JobOptions) ([]*Job, error) {
 		if options != nil && options.ComputeStatus {
 			job.Status = job.GetStatus()
 		}
+
+		// Filter jobs based on tags
+		if options != nil {
```
kevynhale: Since we are already looping on each Result, this shouldn't decrease performance much.

dkron/api.go
```diff
@@ -122,7 +122,22 @@ func (h *HTTPTransport) indexHandler(c *gin.Context) {
...
 func (h *HTTPTransport) jobsHandler(c *gin.Context) {
-	jobs, err := h.agent.Store.GetJobs(&JobOptions{ComputeStatus: true})
+	// Init the Job object with defaults
+	jobOptions := JobOptions{
+		ComputeStatus: true,
+	}
+
+	// Only parse values if there is a body.
+	// If there is not a body, retrieve all jobs.
+	if c.Request.ContentLength > 0 {
```
kevynhale: This makes it so we can still make requests without any tags selected.

victorcoder:
I find this interesting, tell me more about the use case, how many jobs do you have in total?

What's the purpose of the export?

It should be trivial to script this with jq

kevynhale:
Thanks for the response, @victorcoder.

I am currently prototyping dkron to work as a common scheduler for several applications in a large infrastructure. So, currently there are 0 jobs but potentially thousands. Being able to query for a list of jobs and filter on tags is an essential missing piece for a couple different reasons. Different applications will be consuming it and an internal user needs to be able to get a list of jobs for their application only. A more immediate use case is a public facing feature, scheduled reports. For a example a user would need to be able to create a report schedule, edit it, and view a list of only that users reports for that particular application. SinceI am displaying the list of jobs in a UI, it needs to be performant as possible. jq is definitely out of the question, and of course I could filter the list post dkron but that comes at the cost of extra latency and performance.

victorcoder:
@kevynhale I can understand your motivations but TBH it's a feature that I don't want to include right now. Specially with your use case, where you need a high level of confidence that this works well to not show your users something that they don't are allowed to see.

Code is a contract and I can not offer this feature with guarantees in the future, I'm a single developer and I want dkron to provide a level of confidence.

I mean, you have developed the filter yourself in the Dkron and you trust your work and that's ok, you will be the main user of this feature, probably the only one, but it's better for your app security to implement this as a proxy in your application where you can ensure and filter the jobs ensuring that the output will always be correct.

I'm sad to reject good work like this, I can consider including it if there is enough demand in the future.

kevynhale:
Thanks again for the quick response! I understand your reasoning and yes without this change we will have to filter the list after querying dkron. I appreciate your help looking into the feasibility of the PR, and no worries in regards to loss of work as it was more of a POC. Cheers on the great project and keep up the great work!

victorcoder:
@kevynhale could you provide an example request with tag filtering?

victorcoder:
Follow up #412

PR #410, opened 2018-08-01
add create & update job features in web page, use codemirror for JSON editing

wd1900:
Add the abilities to update and create the job in job page.
Remove the initHighlightingOnLoad in dashboard.js, because it is not in use.
image

image


victorcoder:
@wd1900 this is super cool, thanks! Will be even better using https://codemirror.net/ for JSON editing.

wd1900:
@victorcoder yes, you are right, I will add codemirror later.

victorcoder:
LGTM thanks!

PR #412, opened 2018-08-02
Filter jobs by tags

digitalcrab:
Hello guys, 

let me introduce this little change which could be useful for somebody else.

From now on we can assign labels to the jobs. And query (`GET /jobs`) them with labels filter. This change could be useful for the users who have to build UI and display background jobs on their own. One of the use case when you have a different kind of jobs like the system jobs and some application related and you don't want the app to see the jobs not assigned to it.

Exactly our case is that we have 3 types of jobs and 3 different UIs to display the jobs and their status. All jobs should now know about each other. Moreover, we have that 3 UI build for 12 countries. Which lead to 36 UIs. Before that change, we used prefixes like `country-app_name-actual_name_of_the_cron` and it's getting annoying now.

Feel free to comment and accept ;) 

Good luck

-----
UPDATE: 2018-08-29

No more labels, use tags to filter jobs.

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -6,9 +6,9 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/sirupsen/logrus"
 	"github.com/abronan/valkeyrie/store"
-	gin "github.com/gin-gonic/gin"
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -122,7 +122,9 @@ func (h *HTTPTransport) indexHandler(c *gin.Context) {
 }
 
 func (h *HTTPTransport) jobsHandler(c *gin.Context) {
-	jobs, err := h.agent.Store.GetJobs(&JobOptions{ComputeStatus: true})
+	jobTags := c.QueryMap("tags")
+
+	jobs, err := h.agent.Store.GetJobs(&JobOptions{ComputeStatus: true, Tags: jobTags})
 	if err != nil {
 		log.WithError(err).Error("api: Unable to get jobs, store not reachable.")
 		return
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -6,8 +6,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/sirupsen/logrus"
 	"github.com/abronan/valkeyrie/store"
+	"github.com/sirupsen/logrus"
 	"github.com/victorcoder/dkron/proto"
 )
 
--- a/dkron/store.go
+++ b/dkron/store.go
@@ -6,13 +6,13 @@ import (
 	"sort"
 	"time"
 
-	"github.com/sirupsen/logrus"
 	"github.com/abronan/valkeyrie"
 	"github.com/abronan/valkeyrie/store"
 	"github.com/abronan/valkeyrie/store/consul"
 	"github.com/abronan/valkeyrie/store/etcd/v2"
 	"github.com/abronan/valkeyrie/store/redis"
 	"github.com/abronan/valkeyrie/store/zookeeper"
+	"github.com/sirupsen/logrus"
 	"github.com/victorcoder/dkron/cron"
 )
 
@@ -45,6 +45,7 @@ type Store struct {
 
 type JobOptions struct {
 	ComputeStatus bool
+	Tags          map[string]string `json:"tags"`
 }
 
 func init() {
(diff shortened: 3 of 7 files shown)


victorcoder:
Thanks @digitalcrab, this means clearly a vote for https://github.com/victorcoder/dkron/pull/408 that allows the same functionality.

I inclined towards reopening the said PR because it uses the existing job tags. I think the concept of Tags is better suited to filtering elements.

Maybe a mix of both PRs if possible because I like the usage of the query string for filtering.

What do you think?

digitalcrab:
@victorcoder  Tags don't really work because if you tag the job and there are not servers tagged with the same tags this job will not be executed at all. I saw that PR as well, but still I think tags and labels are different in terms of use cases. 

How I see tags ... Bayme you do the same ;) You have a bunch of servers with a different kind of software on it (like we do) one is suitable to run PHP stuff another one backups with Percona tools. We tag servers as "php" and "backup" and we tag jobs to run on certain servers. 

Labels are for logical organization of the jobs, or for grouping them.
Maybe that PR actually wanted labels, but not having them decided to utilize tags. 

@victorcoder  @kevynhale what do you think guys?

kevynhale:
@digitalcrab @victorcoder Sorry for the delay, I agree, when I originally did the PR I hadn't dug into the mechanics of serf and misunderstood the meaning of tags. Having done so now, labels is exactly what we are missing.

I'd like to see labels more as a `map<string, string>` just like tags are, and then add the functionality to filter the `GET /jobs` api on those labels.

an example request might be `curl -X GET  --data '{"label" {"string": "string"}}' http://0.0.0.0:32847/v1/job` - which would only return jobs with the matching label.

so yeah, a combination of both PRs is the way to go.

digitalcrab:
Good evening, @victorcoder 
I am very open to making the required changes. Would you help me and guide in the right direction? ;)
Have a great day

victorcoder:
@digitalcrab @kevynhale I see job tags as AWS tags for resources, are you used to that? Is the way to filter and work with resources from external tools. Some tags could be used by AWS to indicate things (Name) and some other tags are simply free for the user to classify those resources.

You can add an arbitrary number of tags to a job, some of them could be used to filter the nodes on where to run the job and some others could be user for anything you want, like filtering from a third party tool.

IHMO I think that give the user the ability to query defined jobs using tags will give more flexibility, because it can be used to filter for "special" tags to classify jobs and also to filter jobs using already defined tags used for execution filtering. Giving existing users more benefit than adding a new property the need to fill basically providing a way to classify jobs that we already have with Tags.

A @kevynhale proposed in his first PR the interface would be `curl -X GET --data '{"tags" {"string": "string"}}' http://0.0.0.0:32847/v1/job`

Does this makes sense?

digitalcrab:
To me, it doesn't make sense. `GET` with the body? Sounds a bit unusual ;) Even if you check cURL documentation you will see there 

```
 -d, --data DATA     HTTP POST data (H)
     --data-raw DATA  HTTP POST data, '@' allowed (H)
     --data-ascii DATA  HTTP POST ASCII data (H)
     --data-binary DATA  HTTP POST binary data (H)
     --data-urlencode DATA  HTTP POST data url encoded (H)
```

What if I change my PR to use tags for filtering by using a query string and remove labels?

Which approach do you prefer?

1. /?tags=tag1,tag2
2. /?tags=tag1&tags=tag2
3. /?tags[]=tag1&tags[]=tag2

From Go's point of view, the second and third option are equal and implemented in my PR.\
Thank you guys for live discussion ;)

kevynhale:
@victorcoder This makes sense and works for me :+1:

@digitalcrab The 3 options do not provide the same functionality. You are sending a List of strings in the query string params, what we need is a `Map<string, string>` - as that is the type of Tags.

I understand your concern about a body in a `GET` request, though its the easiest way to have a request with a map. There are some frameworks that don't allow a body on a `GET` request, but that is not the case in go's http framework - and it isn't against the rfc.

https://tools.ietf.org/html/rfc7231#section-4.3.1

`A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.`

As in the RFC doesn't care if you do it, just lets you know some implementations don't support it.

digitalcrab:
Ah sorry, I did not pay attention to `map`. Yes. Then you are right. In this case, I am 👍 for the @kevynhale's PR

victorcoder:
Sure @digitalcrab, I mean if using Tags made sense, and it's true that sending a body with a GET request is not usual, actually curl always do a POST when the body is included as shown in the help, but we're forcing a GET with body.

OK that RFC allows it, but I don't like the `some existing implementations to reject the request` I will go with: `?tags[foo]=val&tags[bar]=val`

digitalcrab:
Deal ;) I'll back in 1 day and push the update

digitalcrab:
@victorcoder 
In order to complete this pull request, we need this change from `gin` framework.
https://github.com/gin-gonic/gin/pull/1383
It's merged 3 days ago in master.
**Would you consider updating `gin` to the latest master?**

I don't think that they gonna release it soon. If you check their releases page it happens quite rare.
Your opinion?

victorcoder:
@digitalcrab yes we need this feature in order to make it work. it should work with the latest gin version, "famous last words" 😄

kevynhale:
@victorcoder @digitalcrab Looks like [Gin was released](https://github.com/gin-gonic/gin/releases)! That is some really good timing.

digitalcrab:
Wow! thanks, i'll commit the code later today, then.

victorcoder:
Please merge or rebase with master

digitalcrab:
@victorcoder done ;) sorry for the delay.

PR #419, opened 2018-08-18
Add builtin http plugin

hectorqin:


hectorqin:
Usage:
`
"executor": "http",
"executor_config": {
    "method": "GET",             // Request method in uppercase
    "url": "http://example.com", // Request url
    "headers": "[]"              // Json string, such as "[\"Content-Type: application/json\"]"
    "body": "",                  // POST body
    "timeout": "30",             // Request timeout, unit seconds
    "expectCode": "200",         // Expect response code, such as 200,206
    "expectBody": "",            // Expect response body, support regexp, such as /success/
    "debug": "true"              // Debug option, will log everything when this option is not empty
}
`

PR #420, opened 2018-08-22
Use cobra for commands

victorcoder:
Refactor command line code to use cobra and pflag, also viper to unmarshal config more cleanly. This brings a breaking change in how flags are parsed, they now need double dash `--`. Changes in config style also apply, added some code to support old style config.

Also remove some deprecated command line params.

Use go modules in Go 1.11 working well so far.


dkron/config.go
```diff
@@ -90,39 +90,39 @@ func ConfigFlagSet() *flag.FlagSet {
...
-	cmdFlags.String("bind-addr", c.BindAddr, "address to bind listeners to")
-	cmdFlags.String("advertise-addr", "", "address to advertise to other nodes")
-	cmdFlags.String("http-addr", c.HTTPAddr, "HTTP address")
-	cmdFlags.String("discover", c.Discover, "mDNS discovery name")
+	cmdFlags.Bool("server", false, "This node is running in server mode.")
+	cmdFlags.String("node-name", c.NodeName, "Name of this node. Must be unique in the cluster.")
+	cmdFlags.String("bind-addr", c.BindAddr, "Address to bind network listeners to.")
+	cmdFlags.String("advertise-addr", "", "Address used to advertise to other nodes in the cluster. By default, the bind address is advertised.")
+	cmdFlags.String("http-addr", c.HTTPAddr, "Address to bind the UI web server to. Only used when server.")
+	cmdFlags.String("discover", c.Discover, "A cluster name used to discovery peers. On networks that support multicast, this can be used to have peers join each other without an explicit join.")
```
ProPheT777: I think you can remove or comment it as the feature is no developed, it's missleading

PR #427, opened 2018-09-06
Add executor shell su option

tengattack:
As we may need to change the exec command user and group.
I add this option to dkron-executor-shell

Thanks!


victorcoder:
Thanks for the work! Please, rebase with master to fix testing

PR #429, opened 2018-09-08
Correct test target in Makefile, and make .travis.yml use it

jameshiew:
So that Travis tests that the Makefile target is correct as well


victorcoder:
Thanks!

PR #438, opened 2018-09-13
Update executor support error log instead of rpc error only

tengattack:
Fix executor run command failed but only rpc error status return & store.


tengattack:
Previous log like this:
```
rpc error: code = Unknown desc = exit status 1
```

After patching this PR, we have detail log:
```
exit status 1
firstline
```

tengattack:
Do you have a time to take a look? @victorcoder 
I think the error logs would be the key to identified the issue came from.

victorcoder:
Thanks @tengattack I was doubtful about doing this in the first place, but I like this implementation.
Could you merge or rebase master here and generate the proto file just to make sure it's in sync?

tengattack:
Yep

Issue #390, opened 2018-06-17
Cross-Region (Cross-Data-Center) Resilience Issues - Proposal for a Fix


inadarei:
Current implementation of DKron is only as resilient as the setup of the backend databases it uses. DKron documentation states: "it has no single points of failure due to the use of the fault tolerant distributed databases", but the problem is - these backends (zookeeper, etcd, consul) are only resilient within a single data-center (or a single region, in case of AWS). 

Bottomline: if you set up DKron on AWS and East region has an outage (which happens at least once a year) - you lose DKron.

You are assuming that this issue should be fixed on zookeeper, etcd and consul side and it's not really DKron's problem. Well, that assumption is problematic because as it happens – most large installations of these back-ends solve cross-data-center replication on the client side. You can see Kishore Gopalakrishna, Architect at LinkedIn commenting about it at Quora: https://www.quora.com/Has-anyone-deployed-a-ZooKeeper-ensemble-across-data-centers/answer/Kishore-Gopalakrishna The question/answer is several years old, but I believe the solution is still the same. AFAIK only Consul has automatic cross-region replication, but I believe it is only available in their enterprise version.

The actual common solution to the problem is: client should read only from one region/data-center at any given time but should write to all regions (at least 2) so that if any region/data-center is lost, data is available in other regions and those regions can now become active. 

Here's the proposal for a fix that would allow DKron to support this pattern and dramatically improve resilience of its installations, while introducing a backwards-compatible support:

1. You already have the `-backend-machine` configuration property. Let this be the list of servers that are used for both reading and writing

2. Add new options `-backed-machine-replica` in configuration. DKron would write everything it writes to `-backend-machine` to this server(s) as well, but will not use these servers for reading. 

This very small change will make DKron act in a way compatible with the pattern that is used for deploying cross-region, cross-data-center zookeeper/etcd/consul and will make a lot of people's lives easier.

Thoughts?

victorcoder:
Thanks for the proposal @inadarei, you are right, Dkron is only resilient as it's backend database.

There are several missconceptions in your assumtions. AWS regions are not a single datacenter, Availability Zones are. I've been working several years within a single region using 3 AZs and didn't suffered from and outage in all 3 AZs. This is why for example RDS has the concept of MultiAZ, to ensure resiliency.

I don't know if you can point to some resource that indicates numbers on AWS full regions with all AZs failing. I think that AWS has gone down crossregions globally more than all AZs in one region.

[...]

PR #441, opened 2018-09-17
add support for etcdv3

kevynhale:
In a previous issue about cross data center resilience (https://github.com/victorcoder/dkron/issues/390) you made mention about using mirror-maker with etcd to accomplish cross data center support on the backend.

However, mirror-maker is only supported with etcdv3. The images you are using are running etcdv3, however the client is v2 so you lose the extra functionality with v3. This change gives you the options to use the v3 client with the option `--backend=etcdv3` which is already supported by valkerie.

In order to this to compile you have to use grpc version 1.5.0.


victorcoder:
❤️  @kevynhale

PR #470, opened 2018-12-06
add backend-password for store backend auth

ccl0326:


dkron/agent.go
```diff
@@ -297,6 +297,9 @@ func (a *Agent) StartServer() {
 		if a.config.Backend == "boltdb" {
 			sConfig = &store.Config{Bucket: a.config.Keyspace}
 		}
+		if a.config.Backend == "redis" {
```
victorcoder: Better to refactor this to a switch-case

dkron/config.go
```diff
@@ -73,6 +74,7 @@ func DefaultConfig() *Config {
 		BindAddr:          fmt.Sprintf("0.0.0.0:%d", DefaultBindPort),
 		HTTPAddr:          ":8080",
 		Backend:           "boltdb",
+		BackendPassword:   "",
```
victorcoder: If this is empty string no need to add it to the default values.

dkron/config.go
```diff
@@ -94,6 +96,7 @@ func ConfigFlagSet() *flag.FlagSet {
 	cmdFlags.String("advertise-addr", "", "Address used to advertise to other nodes in the cluster. By default, the bind address is advertised.")
 	cmdFlags.String("http-addr", c.HTTPAddr, "Address to bind the UI web server to. Only used when server.")
 	cmdFlags.String("backend", c.Backend, "store backend")
+	cmdFlags.String("backend-password", c.BackendPassword, "Password to auth for store backend")
```
victorcoder: Just set it to ""

victorcoder: And specify in the comment that this only applies to redis

victorcoder:
Superseded by https://github.com/victorcoder/dkron/pull/505

PR #506, opened 2019-03-19
fix: binary file can't be executed in container

cobolbaby:
When executing './dkron', I encountered an exception 'No such file or directory', but the binary file 'dkron ' really exists.

Relation Ref: https://serverfault.com/questions/883625/alpine-shell-cant-find-file-in-docker


victorcoder:
What problem are you experiencing with the dockerhub docker image?

Does `docker run dkron/dkron` works for you?

cobolbaby:
No, I just encounter the problem when building the docker image with `Dockerfile.hub`

victorcoder:
The `Dockerfile.hub` is intended to be used only by goreleaser, that is launched inside the build container see: https://github.com/victorcoder/dkron/blob/master/Makefile#L8

It's a controlled build environment that ends with the proper working build of the binary, using it from your local environment could end up in unpredictable results.

cobolbaby:
I understand. But the question is about base docker image 'alpine:latest'

I successed to build the custom image just like `dkron/dkron` directory structure. When I run the container, the problem above occur. After searching the causes of the question, I realize that the alpine image lead to the consequence. So I modify the dockerfile and make e it.

cobolbaby:
By pull request, I aim to pursue better compatibility.

victorcoder:
I understand @cobolbaby and I appreciate the contribution don't get me wrong. Maybe this is a misunderstanding, let me put you in context:

1. We know that alpine doesn't use libc, you are compiling with libc so the resulting binary doesn't work on alpine unless your dep is added.

2. Dkron is compiled with goreleaser under a controlled environment explicitly disabling CGO as a choice, see: https://github.com/victorcoder/dkron/blob/master/.goreleaser.yml#L81 to not depend on libc on any linux variant.

3. I prefer this way to have the maximum compatibility and portability for the binary, so no need to add the libc compat dep.

Hope this makes sense. Thanks!

cobolbaby:
Sorry, I ignore `CGO_ENABLED=0` in the .goreleaser.yml before. 

Please forgive me.

PR #524, opened 2019-04-29
Update swagger spec: fix executor_config, add status

yvanoers:
The swagger spec specified the executor_config field of job as an executor_config object that was a map. This change specifies the executor_config field as a map directly. This made the getter in the PHP code generated by swagger work.

Added the status field of job to swagger spec.

Improved text in the description.


victorcoder:
Thanks @yvanoers, what's the problem with the PHP code generated by swagger? is it the official swagger-codegen?

yvanoers:
Yes, the official, unmodified swagger-codegen was used to generate the client.

The problem is/was that the Job object returned by the JobsApi has no value for the ExecutorConfig property. The property is there, but it has no value, even though the API returns one.
This, in turn, caused the getExecutorConfig() method to return an empty value.

What's interesting is that setting a value with setExecutorConfig() and submitting it to the API _did_ properly commit the value in Dkron.

Anyway, IMHO, this PR makes the representation of executor_config in the swagger file match the definition in Go more closely: The ExecutonPluginConfig is defined as being a string map, as opposed to a struct of some sorts.

PR #532, opened 2019-05-14
Add --server option for ubuntu

kakakikikeke-fork:
I couldn't access to localhost:8080 when I started a command `systemctl start dkron`. I think that It's better to have this options.


victorcoder:
I want all nodes to be non servers by default, you need to set `server: true` in the config file located in `/etc/dkron/dkron.yml`

Thx!

Issue #435, opened 2018-09-11
github.com/victorcoder/dkron/plugin package could pull in less dependencies


jameshiew:
Currently the `github.com/victorcoder/dkron/plugin` uses some types and functions from the `github.com/victorcoder/dkron/dkron` package. The `github.com/victorcoder/dkron/dkron` package has many dependencies on big packages like Serf and Etcd which means the dependency graph for any package using `github.com/victorcoder/dkron/plugin` is extremely big. It would be useful if the subset of the `github.com/victorcoder/dkron/dkron` package which is depended on by `github.com/victorcoder/dkron/plugin` was separated out into its own package (maybe `github.com/victorcoder/dkron/types`?)

victorcoder:
Sure this is something to improve, thanks for pointing it.

victorcoder:
done

PR #537, opened 2019-05-20
Plugin Dependency Cleanup

fopina:
After seeing #435 I thought it'd be good to have an idea of the impact of the "plugin dependency cleanup".

### Size changes

* Before patch

```bash
➜  bin ls -lh
total 345104
-rwxr-xr-x  1 fopina  staff    31M May 20 23:16 dkron
-rwxr-xr-x  1 fopina  staff    27M May 20 23:16 dkron-executor-http
-rwxr-xr-x  1 fopina  staff    28M May 20 23:16 dkron-executor-shell
-rwxr-xr-x  1 fopina  staff    27M May 20 23:16 dkron-processor-files
-rwxr-xr-x  1 fopina  staff    27M May 20 23:16 dkron-processor-log
-rwxr-xr-x  1 fopina  staff    27M May 20 23:16 dkron-processor-syslog
```

* After patch

```bash
➜  bin ls -lh
total 176176
-rwxr-xr-x  1 fopina  staff    31M May 20 23:15 dkron
-rwxr-xr-x  1 fopina  staff    11M May 20 23:15 dkron-executor-http
-rwxr-xr-x  1 fopina  staff    11M May 20 23:15 dkron-executor-shell
-rwxr-xr-x  1 fopina  staff    11M May 20 23:15 dkron-processor-files
-rwxr-xr-x  1 fopina  staff    11M May 20 23:15 dkron-processor-log
-rwxr-xr-x  1 fopina  staff    11M May 20 23:15 dkron-processor-syslog
```

*NOTE* I assume changing executor.proto (and executor.pb.go) affects all plugin builds, at least because method signatures changed. I guess those could be kept (with `dkron.` prefix) if just change the package of the file manually (instead of re-generating) though doesn't seem very good idea, maintenance-wise...


fopina:
@victorcoder is this not of interest at all or just on hold until more changes to proto pile up (in order to reduce breaking-changes releases)?

victorcoder:
@fopina sorry for the late response, this is on hold while finishing v2. We need to review this after v2 release, but this is indeed necessary. Thank you.

fopina:
Instead of waiting for the v2 release, should I remake this PR for the v2 branch now?
As it is in beta, might as well get the breaking changes before stable release :)

victorcoder:
@fopina I will do this modification after v2 release. It is a big change in the internals v2 already has big changes. I understand your proposition but with this change the user API won't get modified so could be perfectly done in v2.1 for example.

fopina:
Ah cool, I guess I can close this one then :) cheers!

fopina:
hey @victorcoder would it make sense to re-do this PR with latest master? I see the binaries are now even bigger :)

victorcoder:
Makes sense, though I'm investigating other options on this, but go ahead.

PR #538, opened 2019-05-27
support for authentication with etcd backend

fopina:
`BackendUsername` config flag and added ETCD to the stores using `BackendPassword`


fopina:
knock knock? :)

victorcoder:
thx!

PR #539, opened 2019-05-28
backend TLS support

fopina:
As valkeyrie supports TLS for etcd and consul, dkron could expose that option too.

This is a follow up from #538 but I split into a new PR as I was not sure about this feature being as clean, due to not supporting passing all the actual TLS parameters... but I think supporting all those will add too much jibberish to flags with little gain.

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -1,6 +1,7 @@
 package dkron
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"expvar"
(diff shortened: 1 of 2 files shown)


victorcoder:
Agree, does this works without actually passing TLS params?

fopina:
Yes, I wouldn’t send a PR without minimal testing :) I’m actually using a build with this at work, so far so good. Not sure if there should be a flag for ignore ssl cert but that shouldn’t be used anyway :P

Issue #519, opened 2019-04-25
Cannot use uppercase letters or symbols in job names


thavlik:
## Expected Behavior
I should be able to use any utf-8 characters in a job name, and encodeURIComponent/decodeURIComponent are used when interfacing with the API.

## Actual Behavior
A name like `#HeLLo@@WorLd$` becomes `helloworld`, and you will not be able to delete the job using the same name passed into the DELETE endpoint. If dkron insists on doing things this way, expected behavior entail DELETE transforming the job name just as when it is posted.

The following jest test code passes, demonstrating the error:

```

describe("dkron", () => {
    it("does not work with characters or uppercase", async done => {
        const dkron = Axios.create({
            baseURL: "http://localhost:8080/v1/jobs",
            method: "POST",
        });

        const name = "#HeLLo@@WorLd$"

        console.log(`name = ${name}`);

        // Create a job with that name
        const res = await dkron.post("/", {
            name,
            schedule: `@every 10s`,
            owner: "myself",
            disable: false,
            tags: {
                role: "web:1"
            },
            executor: "http",
            retries: 5,
            executor_config: {
                method: "GET",
                url: `http://localhost:4040/${encodeURIComponent(name)}/`,
                headers: "[]",
                body: "",
                timeout: "30",
                expectCode: "200",
                expectBody: "",
                debug: "true"
            }
        });
        expect(res.status).toBe(201);
        let err: any;
        try {
            await dkron.delete(`/${encodeURIComponent(name)}`);
        } catch (_err) {
            err = _err;
        }
        expect(err).toBeTruthy();
        expect(err.message).toBe("Request failed with status code 404");
        
        done();
    });
});
```

## Steps to Reproduce the Problem

  1. Post a job name with uppercase and/or non-alphanumeric characters
  2. View the name of the job in the web console 
  3. Try and fail to delete the job using the job name you provided dkron with

## Specifications

  - Version: 
```
>  dkron version
Dkron 1.2.2
Agent Protocol: 5 (Understands back to: 2)
```
Platform: Debian 9 (Stretch) AMD64
Backend store: BoltDB

thavlik:
As a workaround, you can sanitize the job name sent to both endpoints using:
```
jobName = jobName.toLowerCase().replace(/\W/g, "");
```

victorcoder:
Hi @thavlik, job names are sanitized before storing it, this means that the original name is changed, and the actual job name won't be the original provided one anymore.

[...]

PR #550, opened 2019-06-18
Reject invalid job names

yvanoers:
Job names with illegal characters are rejected with a 400 HTTP response.
Removed job name sanitation code.
Fixed a typo.

#519


yvanoers:
I started out using validator.v8 as described in the gin/gonic documentation, but that led to much messy code. So then I wrote it this way, much more straight forward and less code. At least I think so.

dkron/api.go
```diff
@@ -248,3 +256,13 @@ func (h *HTTPTransport) jobToggleHandler(c *gin.Context) {
 	c.Header("Location", c.Request.RequestURI)
 	renderJSON(c, http.StatusOK, job)
 }
+
+// isSlug determines whether the given string is a proper value to be used as
+// key for in any backend store (a "slug"). If false, the 2nd return value
```
victorcoder: We're only using one store now in v2.

```suggestion
// key for the backend store (a "slug"). If false, the 2nd return value
```

dkron/api_test.go
```diff
@@ -134,3 +132,35 @@ func TestAPIJobCreateUpdateParentJob_NoParent(t *testing.T) {
 	// Send a shutdown request
 	a.Stop()
 }
+
+func TestAPIJobCreateUpdateValidationBadName(t *testing.T) {
+	a := setupAPITest(t)
+
+	jsonStr := []byte(`{"name": "BAD JOB NAME!", "schedule": "@every 1m", "executor": "shell", "executor_config": {"command": "date"}, "disabled": true}`)
```
victorcoder: Better readability in multiline

dkron/api_test.go
```diff
@@ -134,3 +132,35 @@ func TestAPIJobCreateUpdateParentJob_NoParent(t *testing.T) {
...
+	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
+
+	// Send a shutdown request
+	a.Stop()
+}
+
+func TestAPIJobCreateUpdateValidationValidName(t *testing.T) {
+	a := setupAPITest(t)
+
+	jsonStr := []byte(`{"name": "abcdefghijklmnopqrstuvwxyz0123456789-_ßñëäïüøüáéíóýćàèìòùâêîôûæšłç", "schedule": "@every 1m", "executor": "shell", "executor_config": {"command": "date"}, "disabled": true}`)
```
victorcoder: Same here, better formatting

dkron/api.go
yvanoers: Done.
v2 uses only one store? Which one?

yvanoers: Found it: BadgerDB.

yvanoers:
Conflicts are now resolved and the tests updated to work with the updated code.
Note: I accidentally committed two changes (26b1c07 and 45f9b5c) that weren't supposed to be committed to my v2.x branch. Commit b398477 reverts 26b1c07, but I left the other one in because it only has text corrections.

PR #552, opened 2019-06-19
Fix job not being removed from parent's dependent_jobs when job is deleted

### LINKED ISSUE #499: dependent_jobs not updated
## Expected Behavior

After removing a child job, dependent_jobs of the parent would reflect the new list

## Actual Behavior

After removing a child job, dependent_jobs of the parent still contain the deleted child job, resulting in a post-execution error of the parent job. If there are more children, they might or not be triggered depending if the non-existent child is sent to queue first (and fails).

  1. Create job "a"
  1. Create job "b" with parent_job "a"
  1. Delete job "b"
  1. Check job "a" dependent_job

## Specifications

  - Version: 1.10
  - Platform: linux
  - Backend store: etcd

yvanoers:
Also contains a comment update in the otherwise unrelated cron.go.

Please let me know if you'd like to have this ported to the v2.x branch as well.

Fixes #499.

### DIFF
--- a/cron/cron.go
+++ b/cron/cron.go
@@ -1,5 +1,4 @@
-// This library implements a cron spec parser and runner.  See the README for
-// more details.
+// This library implements a cron spec runner. See the README for more details.
 package cron
 
 import (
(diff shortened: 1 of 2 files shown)


yvanoers:
Replacing PR with a better one.

Issue #526, opened 2019-05-02
Feature request: "Start/End Time" displayed in configured timezone


victorcoder:
Our server runs in the Time zone Europe/Amsterdam (CEST, +0200) and we configure our cron jobs with   "timezone": "Europe/Amsterdam"
Only the “Start Time” and “End Time” within the email execution report are displayed in UTC time. 
 
I would be nice if the execution report would display the “Start Time” and “End Time” in the time zone configured in the job. So in our case “Europe/Amsterdam”.

victorcoder:
Fixed

PR #557, opened 2019-06-26
Represent execution times in the job's timezone

yvanoers:
Implements #526.
This applies the job's timezone to the execution's start and end times when executions get unmarshalled, so everything that uses them afterwards, including the dashboard and the notifier should present the times in the job's timezone.

@victorcoder, let me know what you think.


dkron/store.go
```diff
@@ -627,7 +627,16 @@ func (s *Store) Shutdown() error {
 	return s.db.Close()
 }
 
-func (s *Store) unmarshalExecutions(items []*kv, stopWord string) ([]*Execution, error) {
+func (s *Store) unmarshalExecutions(items []*kv, jobName string) ([]*Execution, error) {
+	// Get the job's timezone
+	var jobLocation *time.Location
+	if job, err := s.GetJob(jobName, nil); err == nil {
```
victorcoder: Could we avoid going to the store while unmarshaling?

yvanoers: I'm not too thrilled about that either. Avoiding it might require a lot of plumbing - but I'll see what can be done about it.

yvanoers: How about this:
Commit 1d614c3 refactors things so the timezone gets passed down to unmarshalExecutions.
Now, only the dashboard goes to the store to retrieve the timezone.

victorcoder:
This diff is also not ok, a clean up is needed.

PR #558, opened 2019-06-29
Fix dependent job updates in parent jobs on change or delete (v2.x)

### LINKED ISSUE #499: dependent_jobs not updated

yvanoers:
This PR includes several tests for for this functionality.
FIxes #499.


yvanoers:
While I was testing this I noticed that jobs can be deleted when they have dependent jobs. If that happens, the child's `parent_job` does not get altered.
If a parent job having the same name is then created again, the parent will not have any dependent_jobs. The jobs will be in an inconsistent state.

@victorcoder What would be the correct behavior in this scenario according to you? Restricting deletions of parents, recovering dependent_job relationship at creation, removing the parent job from dependent jobs, ... ?

My preference would be to restrict deletion of a jobs having dependent jobs. Not doing so would require a lot of tricky edge cases to be solved.
On the other hand, restricting deletions would put the burden of deleting jobs the right way on the users.

victorcoder:
@yvanoers I rebased v2.x branch with master, could you rebase too?

Regarding your question, On parent deletion, I would remove the parent_job from childs and that's it. The user should be responsible for removing orphaned jobs or reasigning a parent job.

yvanoers:
@victorcoder I'm trying to get my repos aligned with yours, but I'm having a hard time doing it:
My local repo says it is up to date with upstream (distribworks/v2.x) and origin (yvanoers/v2.x), but github shows differences between the two. I have no clue why that is.
I'm considering removing my fork and forking again just to get the mess sorted out, and then recreating the PRs.
Unless you have some suggestions, that's probably what I'm going to do.

victorcoder:
@yvanoers yup, that would be the best solution, I think when I force pushed with the rebase your fork got unaligned.

yvanoers:
@victorcoder might very well be. Ok, I will refork and recreate the PRs tomorrow, my head hurts now.

Issue #548, opened 2019-06-17
Why load all record in pagination


pMonfared:
Pagination doesn't work correctly in dashboard, why load all record ? get 10 record per page
please fix that!

![image](https://user-images.githubusercontent.com/10448255/59610561-e5b6db00-912e-11e9-98db-d73ffb597830.png)

victorcoder:
@pMonfared What is not working for you?

victorcoder:
Needs to use the issue reporting template.

pMonfared:
pagination :
load 10 record per page, not load all records with one request!
if you get all record with one request , why use pagination?
and i search a title when current page is 1 , don't show filtered record, show nothing but filtered record found on other page!

victorcoder:
> if you get all record with one request , why use pagination?

Modern browsers are perfectly capable of rendering thousands of json objects without problems. The question is, why you shouldn't use pagination? It's unreadable to show thousands of records in the same page. This is why solutions like pagination.js exists.

> and i search a title when current page is 1 , don't show filtered record,

That's right the filter is per page, not the whole set.

fopina:
@victorcoder regardless of the way this was reported, do you consider filtering being per page a bug or was it intended? just wondering as it was something I had planned to do a PR if it's not changed in v2

saeedmsdi:
@victorcoder pagination required on large records. modern browsers cant support 1milion records! and loading this huge records is unuseful. 

and for the filtering : like I need today schedules failed, or next day scheduled. it's very useful for us to find filtered records

thanks

victorcoder:
@fopina filtering per page was intended but doesn't mean it's ideal, the impossibility to dedicate more time it was the most decent solution at the time.

I gladly accept a PR addressing this issues, though I will try to work on it too.

victorcoder:
We are mixing several issues here, let's close this as fixed in #561 and reopen a feature request or PR for filtering.

PR #561, opened 2019-06-30
Fixes paginator gap and add First and Last button

victorcoder:
Though not an ideal paginator but more usable in case of high number of jobs.

Gap fixed to 10 and added First and Last buttons to navigate.

Related to #548


static/js/dashboard.js
```diff
@@ -166,7 +166,11 @@ dkron.controller('JobListCtrl', function ($scope, $http, $interval, hideDelay) {
 
   // calculate page in place
   $scope.groupToPages = function () {
-    $scope.gap = Math.round($scope.jobs.length / $scope.itemsPerPage);
+    if (Math.round($scope.jobs.length / $scope.itemsPerPage) < 10) {
+      $scope.gap = Math.round($scope.jobs.length / $scope.itemsPerPage);
+    } else {
+      $scope.gap = 10;
+    }
```
yvanoers: $scope.gap = Math.round(....
$scope.gap = Math.min($scope.gap, 10);

Saves duplication of the gap calculation and an if.

victorcoder: Thanks for that!

PR #566, opened 2019-07-07
Represent execution times in the job's timezone

yvanoers:
Implements #526, superseeds #557 .

### DIFF
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -256,7 +256,7 @@ func (j *Job) GetStatus() string {
 		return StatusNotSet
 	}
 
-	execs, _ := j.Agent.Store.GetLastExecutionGroup(j.Name)
+	execs, _ := j.Agent.Store.GetLastExecutionGroup(j.Name, j.GetTimeLocation())
 	success := 0
 	failed := 0
 	for _, ex := range execs {
@@ -310,6 +310,14 @@ func (j *Job) GetParent() (*Job, error) {
 	}
 
 	return parentJob, nil
+}
+
+// GetTimeLocation returns the time.Location based on the job's Timezone,
+// the default (UTC) if none is configured or nil if an error occurred
+// while creating the timezone from the property
+func (j *Job) GetTimeLocation() *time.Location {
+	loc, _ := time.LoadLocation(j.Timezone)
+	return loc
 }
 
 // GetNext returns the job's next schedule from now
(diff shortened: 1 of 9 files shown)


dkron/job.go
```diff
@@ -275,6 +275,25 @@ func (j *Job) GetParent() (*Job, error) {
...
+		return DefaultJobTimeLocation()
+	}
+	return loc
+}
+
+// Returns a UTC time.Location
+func DefaultJobTimeLocation() *time.Location {
+	loc, err := time.LoadLocation("UTC")
+	if err != nil {
+		log.Panicf("Failed to get Location instance for UTC!")
```
victorcoder: This expects a format string and arguments.
`Panicf -> Panic`

dkron/job.go
```diff
@@ -275,6 +275,25 @@ func (j *Job) GetParent() (*Job, error) {
...
+func (j *Job) GetTimeLocation() *time.Location {
+	loc, err := time.LoadLocation(j.Timezone)
+	if err != nil {
+		return DefaultJobTimeLocation()
+	}
+	return loc
+}
+
+// Returns a UTC time.Location
+func DefaultJobTimeLocation() *time.Location {
```
victorcoder: I would move this to the more general `utils.go` and refactor its name to `DefaultTimeLocation` as it applies to several entities across the code, what do you think?

yvanoers: `DefaultJobTimeLocation` is always used alongside the job name. The pair is the job context for an execution. That is why I put it in `job.go`.
If job would have been in its own package, I would have named it `DefaultTimeLocation`, and it could have been more appropriately been referred to as `job.DefaultTimeLocation`.

What it is doing is indeed very generic, though. When reasoned about from the function outwards, disregarding what it is supposed to mean, and only looking at what it does, it would indeed best be placed in `utils.go`. But then I suggest renaming it to `UTCTimeLocation` or `GetUTCTimeLocation`.

I can eliminate two uses of `DefaultJobTimeLocation` by passing a job from further down the callstack (eliminates use in `agent.go` and `store.go`). The remaining occurrences in `job.go` and `dashboard.go` can do without function's "job" semantic and it could then happily be moved to `utils.go`.

Shall I give that a try?

yvanoers: I removed DefaultJobTimeLocation altogether.

yvanoers:
Closing this PR as v2.x has now been merged to master. Opened PR #614 to replace this.

Issue #554, opened 2019-06-21
Feature request: Job "display" name


knightsg:
First of all, thanks for developing this. We were looking for something simple but robust to handle our scheduled jobs and Dkron fits the bill perfectly.

One suggestion/idea we've had after the relatively short amount of time we've been using it is that it would be nice if we could specify an alternative display name/label for jobs that will be shown in the UI job list. I realise that the actual job name requires a specific format for use in the k/v store but it would be useful if there was another field you could use to set a more readable name that the UI would display if it was present.

victorcoder:
Thanks @knightsg, TBH this will not be a priority right now but I gladly accept a PR implementing it.

Idea: Use the new metadata property in v2 to display a new column if it has the `name` metadata field.

victorcoder:
Implemented in #568

PR #568, opened 2019-07-09
Add displayname property to jobs

yvanoers:
Implements #554 .
I chose to create a separate displayname field instead of using the metadata. IMO, whatever the user stores in metadata should have no effect whatsoever on the behavior of the application.


yvanoers:
@victorcoder Squashed!

victorcoder:
Thanks!

PR #588, opened 2019-08-22
Fix lookup of non-existent job should return 404

### LINKED ISSUE #586: API returns uninitialized job for non existing job in v2.0.0-b9
Use this template if you are reporting a bug. Don't use this template of you are proposing a feature.

## Expected Behavior
/v1/jobs/<job_id> with a non existing job_id returns status code 404, as in version 1.2.x

## Actual Behavior
Non existing job_id returns status code 200 OK with an uninitialized job in the body. 

  1. Go to /v1/jobs/<job_id> with non existing job ID 

## Specifications

  - Version: 2.0.0-beta9
  - Platform: centOS 7
  - Backend store: default

yvanoers:
Fixes #586.

### DIFF
--- a/dkron/api_test.go
+++ b/dkron/api_test.go
@@ -203,3 +203,15 @@ func TestAPIJobCreateUpdateValidationValidName(t *testing.T) {
 
 	assert.Equal(t, http.StatusCreated, resp.StatusCode)
 }
+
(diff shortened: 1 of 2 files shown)


yvanoers:
@victorcoder 

> Maybe return nil for err

Done!

Issue #578, opened 2019-07-25
Immediately run jobs


alexahnfetch:
Previously mentioned in https://github.com/distribworks/dkron/issues/509, but I'm opening a ticket hoping that I can change your mind. Setting an arbitrary padding to set a future schedule date is not a rigorous solution, and it forces developers to think about all the things that can potentially go wrong (e.g. clock skew) - to the point that they will conservatively use a large padding.

It would be nice if we had something like:
```
@at now
```
or allow an empty `schedule` field:
```
curl localhost:8080/v1/jobs -XPOST -o - -D '{
  "name": "job1",
  "schedule": "",
  "timezone": "Europe/Berlin",
  "owner": "Platform Team",
  "owner_email": "platform@example.com",
  "disabled": false,
  "tags": {
    "dkron_server": "true:1"
  },
  "concurrency": "allow",
  "executor": "shell",
  "executor_config": {
    "command": "date"
  }
}'
```
where an empty `schedule` field specifies an immediately run job (similar to https://mesos.github.io/chronos/docs/api.html#adding-a-scheduled-job).

## Expected Behavior
```
curl -v localhost:8080/v1/jobs -XPOST -d '{
>   "name": "job1",
>   "schedule": "",
>   "disabled": false,
>   "tags": {
>     "dkron_server": "true:1"
>   },
>   "concurrency": "allow",
>   "executor": "shell",
>   "executor_config": {
>     "command": "date"
>   }
> }'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> POST /v1/jobs HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 204
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 214 out of 214 bytes
< HTTP/1.1 201 Created
< Content-Type: application/json; charset=utf-8
< Location: /v1/jobs/job4
< Date: Thu, 25 Jul 2019 00:39:38 GMT
< Content-Length: 427
< 
* Connection #0 to host localhost left intact
[...]

PR #592, opened 2019-08-26
Autorun job after create

yvanoers:
This PR is how I imagine the implementation of a feature allowing a job that gets run when it gets created.

It introduces a `runoncreate` querystring parameter for the create/update API call, and the `@manually` predefined schedule.

Together these allow dkron to be used kind of like a task queue, as discussed in #578. The main difference with an actual task queue is that dkron has no throttling mechanism and would just run tasks as they are created. Dkron does not scale in this regard, so if too many requests are sent, it might drown in jobs. Use with care.

To use:
```
url -X POST -v -o - -d '{
  "name": "job1",
  "schedule": "@manually",
  "executor": "shell",
  "executor_config": {
    "command": "echo Hello world"
  }
}' 'localhost:8080/v1/jobs?runoncreate'
```

### DIFF
--- a/cron/parser.go
+++ b/cron/parser.go
@@ -225,6 +225,9 @@ func parseDescriptor(spec string) Schedule {
 			Month:  all(months),
 			Dow:    all(dow),
 		}
+
+	case "@manually":
+		return At(time.Time{})
 	}
 
 	const every = "@every "
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -165,6 +165,11 @@ func (h *HTTPTransport) jobCreateOrUpdateHandler(c *gin.Context) {
 		return
 	}
 
+	// Immediately run the job if so requested
+	if _, exists := c.GetQuery("runoncreate"); exists {
+		h.agent.GRPCClient.RunJob(job.Name)
+	}
+
 	c.Header("Location", fmt.Sprintf("%s/%s", c.Request.RequestURI, job.Name))
 	renderJSON(c, http.StatusCreated, &job)
 }
--- a/website/content/swagger.yaml
+++ b/website/content/swagger.yaml
@@ -65,6 +65,13 @@ paths:
           required: true
           schema:
             $ref: "#/definitions/job"
+        - in: query
+          name: runoncreate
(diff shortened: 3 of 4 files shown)


victorcoder:
Cool @yvanoers, lean implementation, I like it, thank you.

You are totally right, dkron is not designed as a "queue" in the sense it's not FIFO when running jobs.

Maybe this could be a thing to work on to provide a simple solution, we can discuss it by mail if you want.

victorcoder:
We can go with this after fixing CI

yvanoers:
I added some documentation to go with this.
CI reports A-OK.

fopina:
Awesome!

fopina:
> You are totally right, dkron is not designed as a "queue" in the sense it's not FIFO when running jobs.

@victorcoder how does dkron handle several @hourly jobs with little number of agents (let's say 1)?
are they not queued? they all start at the same time?

victorcoder:
@fopina they all start at the same time, no queues.

fopina:
So it’s the same issue right? Nothing new...

victorcoder:
What do you mean @fopina ?

fopina:
I mean that scheduling a job at creation is not a “new” issue if jobs that are scheduled at same time already run at same time. Kind of following up on the discussion that led to this PR :)
One change to help would probably be the “hashed” @hourly strategy like Jenkins does.
So if I want a job to run every hour but I don’t mind which minute/second, we can use “H” placeholder in the cron syntax and Jenkins replaces that with an hash from job id/name.
This allows daily/hourly jobs to run at different times to reduce peaks

yvanoers:
From my perspective, this is a new feature request, separate from the discussion leading up to this PR.
Maybe it is a good idea to open a new feature request to discuss this?

fopina:
Definitely agree with being new feature request, I was just following up on the no-queue argument and mixed in a feature suggestion, apologies

yvanoers:
No worries :)

victorcoder:
I like the idea of the hashed @hourly, indeed a new feature request. Thanks @fopina

PR #595, opened 2019-09-05
Improve validation of job values, return 4xx error when validation fails

yvanoers:
This especially prevents jobs with an invalid schedule from getting stored.
E.g. a schedule of `@at notatime` is now rejected, instead of stored and afterwards ignored by the scheduler.

Concurrency is also validated, but that is a user courtesy, not a technical necessity.

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -850,7 +850,7 @@ func (a *Agent) RefreshJobStatus(jobName string) {
 }
 
 // applySetJob is a helper method to be called when
-// a job property need to be modifeid from the leader.
+// a job property need to be modified from the leader.
 func (a *Agent) applySetJob(job *proto.Job) error {
 	cmd, err := Encode(SetJobType, job)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


yvanoers:
> duplicating code
Indeed, I didn't spot that.

Then, there's something fishy going on elsewhere. I'm reworking this.

yvanoers:
@victorcoder 
Now all validation code is in the store, and it is called from the api. But I'm not happy with this:
`new (Store).validateJob(&job)`
IMO, that ought to be:
`job.validate()`

The validation code is out of place in the store. Sure the store can validate a job before it is persisted, but the functions `(s *Store) validateJob(job *Job) error` and `(s *Store) validateTimeZone(timezone string) error` never use `s`. I think these functions should go in `job.go` and the store should call them using the job object.
Do you agree?

yvanoers:
@victorcoder I've refactored it as proposed and fixed some other stuff too.
Please have a look, appreciate your feedback!

Issue #596, opened 2019-09-05
Custom TLS certificates for HTTP executor


Demonian:
Is there ability to make add custom certificate to `http-executor` or at least disable TLS verification? Is there a plan to make this possible in future?

## Expected Behavior
Successful call to services that are secured with custom TLS certificates from `http-executor`

## Actual Behavior
Call fails with TLS errors.

## Steps to Reproduce the Problem

  1. Create `http executor` Job with call to service secured with custom TLS certificated
  2. Trigger the Job.

## Specifications

  - Version: v2.0.0-rc1
  - Platform: k8s
  - Backend store: embedded

yvanoers:
@Demonian As far as I can tell, the http-executor has no TLS configuration options.

The recommended way to go about, though, this is to add the CA of your self-signed certificate to the trusted Root CAs of the systems running Dkron.

Whether any TLS configuration options might get included I'll leave for @victorcoder to answer.

victorcoder:
@Demonian IMO an option to ignore untrusted certificates could be added, will be disabled by default and enabled on users demand.

victorcoder:
Fixed in #603

PR #603, opened 2019-09-14
Add TLS configuration options to http executor

yvanoers:
Implements option `tlsNoVerifyPeer` to disable SSL validation (#596)
Implements options `tlsCertificateFile` and `tlsCertificateKeyFile` to specify a client certificate. This will allow Dkron's http executor to authenticate to web sites protected by a client certificate.


website/content/usage/executors/http.md
```diff
@@ -17,6 +17,9 @@ timeout: Request timeout, unit seconds
 expectCode: Expect response code, such as 200,206
 expectBody: Expect response body, support regexp, such as /success/
 debug: Debug option, will log everything when this option is not empty
+tls_noverifypeer: false (default) or true. If true, disables verification of the remote SSL certificate's validity.
```
Demonian: Great PR =) I have only one proposal to make arguments names more canonical (E.g. istio, kiali, openssl etc.):
```
insecure_skip_verify (instead of tls_noverifypeer)
client_cert (instead of tls_certificatefile)
client_key (instead of tls_certificatekeyfile)
```
Do we also need to have ability to configure root CA cert?

yvanoers:
@victorcoder Names are updated. LMK if there's anything more.

website/content/usage/executors/http.md
Demonian: @yvanoers Ok, agree about names=) I was adding custom RootCA to docker image root CA, to have ability to use self signed certificates, so probably it will be useful.

yvanoers: @Demonian Ok, I will see what I can do to add an option for the Root CA.

PR #616, opened 2019-10-27
Remove Storage interface

yvanoers:
Because v2.x moved away from supporting multiple storage engine, the Storage interface might as well be removed.


victorcoder:
I need to keep this because of Pro features, thanks anyway!

PR #618, opened 2019-10-29
Use Go 1.13 for testing

victorcoder:
Upate dockerfile and remove testing dependency


victorcoder:
I don't know why travis broke after upgrading to Badger 2RC4 closing this in favor of a new strategy

Issue #493, opened 2019-02-12
feature: streamable output


fopina:
It'd be really nice to have "streamable output" in the UI/API for the jobs instead of only getting the output in the end.
I was considering trying to make a PR but I got quite lost in the effort to accomplish this as, due to this being a plugin system, a lot of changes would be required just to be able to get pieces of information out of an execution without it being done...

Any hints/workarounds on how I could accomplish this?

victorcoder:
Agree, it will be a nice feature @fopina, as you say, it could be a bit more complex to implement due to the usage of plugins but doable anyway.

I will start investigating on gRPC streaming or just capture the stdout of the plugins and refactor them to progressively send the output to the agent.

fopina:
Awesome. I’m new to go and even newer (if possible) to this plugin structure but do let me know if I can help in some way

fopina:
@victorcoder how would you store the streamed output? same as now (and keep updating that value on every stream update)?
It'd probably make sense to refactor the UI into opening a new window (with auto-refresh) for the output instead of that modal (with pre-loaded content), right?

Asking in case I decide to give it a go myself :)

victorcoder:
@fopina yes, that makes sense!

fopina:
In the meantime would you take a PR to change shell executor printing output at end of command to print as it goes?
That’d allow for the stderr/stdout to serve as an alternative, as currently there’s no way at all to know current output of a running job

victorcoder:
Sure, makes sense

fopina:
I was starting the change yesterday but realized that printing output as it goes will get mixed in stdout with the rest of dkron output. Before it was all together which still looked weird but with this changes it’s random lines in the middle without any info of where that output came from... any suggestions?

victorcoder:
@fopina I have a plan for this, we can discuss it by mail.

fopina:
reached out to the email in your commit history

l-we:
Is there any progress?

victorcoder:
Done in #719 #716

PR #622, opened 2019-11-12
WIP: streamable output

fopina:
Now that v2 has gone to master, I'm trying to make an attempt at #493 

Based on the [official bidirectional example](https://github.com/hashicorp/go-plugin/tree/master/examples/bidirectional), the idea is:

* add `StatusHelper` to `ExecuteRequest`
* `invoke` should then pass (or expose actually) a callback that will receive partial output
* plugins call (or not) the status update callback whenever they feel like they should

Questions:
* try to keep it compatible with already built plugins? as 2.0 already breaks them and is still in RC, I guess this is probably not a priority?
* initial parameters for callback method:
   * `progress float32` - `0 to 1`, representing total progress in case the plugin can report such thing. if doesn't make sense to plugin, use `-1`
   * `output byte[]` - partial output to go into server storage
   *  `error bool` - if it's error output or normal output - just to avoid changing interface in the future, this can be useful to use RED in the logs for error output. `shell` executor would use error=True for Stderr

Does this make sense @victorcoder ? Any design changes/inputs before proceeding?


fopina:
Another approach would be to simply use plugin logging and parse that.
Not sure if it would be more efficient but I went for this one as it looks cleaner

victorcoder:
Hi @fopina thanks for the work, there is some concepts in this PR that I like and that I consider that must be addressed first in a new PR, for example upgrading the generated proto code.

My idea to implement this was a bit different, instead of having a callback/method waiting to receive feedback from the node running the job, I was thinking in establish a long running gRPC connection to the server from the node. This will provide a robust way to detect dead nodes, also will open the door to report the status info of the nodes much better.

Though, I like a lot the bidirectional implementation for the plugin, actually this was the missing piece in the puzzle.

I wanted to go with the RC as it is, without adding more features, but we can include this in the 2.1 as soon it is ready.

Let's start upgrading the generated proto code in a new PR, then work with this PR to implement gRPC streaming in the call to `RunJob`

fopina:
Thanks for the feedback. I'll stop working on this approach then. I'd like to have some sort of streamable output working by end of next week, but I'll just hack something up on the side with this setup until I understand a bit more what you have in mind :) 👍

savvasdalkitsis:
Is this PR abandoned? Having streamable logs for long running tasks would be SO helpful. I never know if my backup tasks have frozen up or are still running :/

fopina:
It would indeed :)
Not really sure where we left this at as @victorcoder had some bigger changes in mind
But I’m happy to pick this up again given some guidance on preferred approach

victorcoder:
sure @fopina and @savvasdalkitsis this is a must, I want to start working on, the approach to take here is "long running gRPC connections", that would help to improve two of the biggest current flaws, Job execution status in realtime and streamable output.

@fopina Let me know if you want to take this or should I start and include you in the loop.

fopina:
I don’t mind taking a swing at it, just need a bit more guidance as not sure where to start with those long connections to the server :)

victorcoder:
@fopina, currently the dialog with an agent is as follows:

1. Job trigger run -> gossip `job:run` to target agents
2. Agent gRPC call server to ask for Job
3. Agent starts execution and when finished calls back server with `ExecutionEnd` returning job output and result.

The new workflow will be:

3. Agent opens a streaming gRPC call to the server on execution start with a new `AgentRun` gRPC method.
4. Start execution while streaming the progress.
5. Closes the connection when execution finishes.

This implementation should be pretty straightforward and will allow for a 100% reliable of job execution status while a job is running, and as a side effect job output streaming in realtime.

victorcoder:
@fopina as #716 is already merged, we can now start working in the bidirectional stuff for plugins, I will copy all what you implemented in this PR and create a new PR.

I'm thinking in not adding the `Progress` property to simplify things and because I don't think it offers much value at this point.

@savvasdalkitsis follow-up the mentioned PRs, this feature will land in 2.2 in the following days

fopina:
This is awesome! I’m not sure how helpful I can be these days as time is quite short but I’ll definitely test it out if you need me to

PR #638, opened 2019-11-19
refactor: Replace boltdb with fastlog

victorcoder:
Raft log was been handled using BoltDB, but the KV store is BadgerDB that is more performant than Bolt. This rendered in a storage speed bottleneck in the Raft log store which always comes first in write operations.

Using https://github.com/tidwall/raft-fastlog eliminates the bottleneck in Raft log storage as it is even faster than BadgerDB.

This is a breaking change when upgrading to > 2.0-RC9


victorcoder:
Closing for now

PR #650, opened 2019-12-01
Backup/Restore jobs with file

vision9527:
issue: https://github.com/distribworks/dkron/issues/212


vision9527:
Test file in scripts/testBackupJobs.json.
It's just backend API.

victorcoder:
@vision9527 I would remove the download part as it doesn't bring any new feature at all, the user could call /jobs and redirect the output to a file for the same effect.

And could you squash to 1 commit? thx!

vision9527:
@victorcoder 
Done. I have removed download api and squash to one commit.

PR #654, opened 2019-12-09
restore jobs with file

vision9527:
restore jobs with file


vision9527:
@victorcoder 
I have removed download api and squash to one commit.

victorcoder:
Thanks for your work @vision9527 are you still interested in this?

vision9527:
> Thanks for your work @vision9527 are you still interested in this?

I'm just in this case. Respect your design philosophy. It's up to you.

dkron/api.go
```diff
@@ -73,6 +75,7 @@ func (h *HTTPTransport) APIRoutes(r *gin.RouterGroup, middleware ...gin.HandlerF
 	v1.PATCH("/jobs", h.jobCreateOrUpdateHandler)
 	// Place fallback routes last
 	v1.GET("/jobs", h.jobsHandler)
+	v1.POST("/restoreJobs", h.restoreJobsHandler)
```
victorcoder: I would call this just `/restore` or `/load`

dkron/api.go
```diff
@@ -212,6 +215,42 @@ func (h *HTTPTransport) jobRunHandler(c *gin.Context) {
 	renderJSON(c, http.StatusOK, job)
 }
 
+// Restore jobs from file.
+// Overwrite job if the job is exist.
+func (h *HTTPTransport) restoreJobsHandler(c *gin.Context) {
```
victorcoder: same for handler `restoreHandler`

victorcoder:
Also modify the `swagger.yml` spec please.

victorcoder:
@vision9527 also you will need to fix conflicts

dkron/api.go
vision9527: done

vision9527:
@victorcoder done

dkron/job.go
```diff
@@ -396,3 +399,68 @@ func isSlug(candidate string) (bool, string) {
...
+		}
+		if isTopParentNodeFlag {
+			j++
+		}
+		jobs = rejobs
+	}
+	return jobs, nil
+}
+
+// find parent job and validate job
```
victorcoder: This is not correctly linted, it should start with `findParentJobAndValidateJob...`

victorcoder:
@vision9527 could you write the rationale for creating a job tree. I see some test jobs without dependent jobs. In theory if the jobs were exported with the `/jobs` endpoint, importing them one by one should end up with a correct structure without having to elaborate such a complex parsing.

vision9527:
> @vision9527 could you write the rationale for creating a job tree. I see some test jobs without dependent jobs. In theory if the jobs were exported with the `/jobs` endpoint, importing them one by one should end up with a correct structure without having to elaborate such a complex parsing.

Oh, it's my mistake. I change the test jobs code. You can review now.  The job tree that has parent job and dependent job can be created with a correct structure. The key is to deal with parent job and dependent job in the endpoint.

victorcoder:
@vision9527 Shouldn't the enpoint in this comment https://github.com/distribworks/dkron/pull/654#discussion_r393615297 be just `restore`

I will ask it in another way, what will happen if we just do:

```go
// file: api.go
...
var jobs []*Job
err = json.Unmarshal(data, &jobs)

for _, job := range jobs {
  err := a.GRPCClient.SetJob(job)
}
...
```

vision9527:
> @vision9527 Shouldn't the enpoint in this comment [#654 (comment)](https://github.com/distribworks/dkron/pull/654#discussion_r393615297) be just `restore`
> 
> I will ask it in another way, what will happen if we just do:
> 
> ```go
> // file: api.go
> ...
> var jobs []*Job
> err = json.Unmarshal(data, &jobs)
> 
> for _, job := range jobs {
>   err := a.GRPCClient.SetJob(job)
> }
> ...
> ```

image
Abort if parent not found before committing job to the store.
It can't be for sure that the create job sequence is right.
Job tree is a array in which parent job is in the front of  the dependent jobs, so it can be sure when  the dependent job will be created that parent job is exist.

vision9527:
@victorcoder Please review, bro.

victorcoder:
Many changes in master, please fix conflicts or rebase, and we should be good to go, thanks!

vision9527:
> Many changes in master, please fix conflicts or rebase, and we should be good to go, thanks!

Done.

dkron/store.go
```diff
@@ -149,6 +149,14 @@ func (s *Store) SetJob(job *Job, copyDependentJobs bool) error {
 		if err := s.removeFromParent(ej); err != nil {
 			return err
 		}
+
+		if ej == nil || ej.ParentJob == "" {
```
victorcoder, line 153: ej can't be nil at this point, no need to check this

dkron/store.go
```diff
@@ -149,6 +149,14 @@ func (s *Store) SetJob(job *Job, copyDependentJobs bool) error {
...
+			// Need to handle this case already created parent job when dependent job is not exist
```
victorcoder, line 154: Please reword this comment as it's not clear what it means.

dkron/store.go
vision9527: yes, you are right. I have clean this.

Issue #655, opened 2019-12-09
Concurrency Forbid being ignored (again) since upgrade to 2.0


geoffjukes:
Hi @victorcoder 

Since upgrading to 2.0 (currently 2.0.1) we have an issue where the `concurrency: forbid` is being ignored after running for some time.

![image](https://user-images.githubusercontent.com/1959570/70476157-fff0c680-1a8a-11ea-886c-002c9cb15783.png)

I'm not sure how to start debugging the issue.

victorcoder:
Thanks for reporting @geoffjukes I will investigate.

geoffjukes:
@victorcoder Were you able to identify an issue? or can I do anything to help debug?

For now I have disabled the affected jobs, and am firing them manually.

victorcoder:
@geoffjukes yes, the feature was broken in https://github.com/distribworks/dkron/pull/630/files#diff-9719285e240fa82e7f8e90770c4df6f0L338 , It's fixed but I'm adding an integration test to ensure it doesn't happen again.

Release coming in the following hours.

geoffjukes:
@victorcoder Awesome, glad I wasn't losing my mind :) Thanks for the prompt response.

PR #658, opened 2019-12-10
refactor: Die on plugin communication error

victorcoder:
All plugins uses safe code for error handling so in this case we can not do anything but die, exactly the same as in processor plugins.

Related to https://github.com/distribworks/dkron/issues/652

### DIFF
--- a/plugin/executor.go
+++ b/plugin/executor.go
@@ -34,7 +34,13 @@ type ExecutorClient struct {
 
 func (m *ExecutorClient) Execute(args *dkron.ExecuteRequest) (*dkron.ExecuteResponse, error) {
 	// This is where the magic conversion to Proto happens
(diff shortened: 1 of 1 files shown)


davidgengenbach:
@victorcoder This will result in the main process getting killed when there was an error executing on the client (= the plugin) - as far as I understand?
This also means that a lot of time can pass before the plugin-death has been detected (= for jobs which are executed every week or something).
It may be interesting to add periodic health-checks to check for the plugin-health - and proceed to die when the plugins are not healthy.

Just an idea!

victorcoder:
@davidgengenbach plugins processes should be considered as being the same thing as the main process, if one dies the main process should die fast as you said.

I will go with this PR better than nothing and I will left #655 open as an improvement.

PR #659, opened 2019-12-11
Validate empty job names

### LINKED ISSUE #656: Dkron UI allows empty job name
Version: 
2.0.1

Problem: 
User is able to create a new dkron job with the name being an empty string. After creating this job, it is not possible delete the job (returns a 404), and it is not possible to update the name. When attempting to create a new job with an empty name, the new job will simply override the old job without warning. Adding a display name will show a value on the jobs list, but clicking on that job display name results in a 404.

Expectation: 
Jobs with an empty string name should not be allowed

Steps to reproduce:

1. Create a basic job with the job name being an empty string and use a valid parent job. (no parent or empty parent field will return an error that it cannot be its own parent)
2. Attempt to delete the newly created job from the jobs list

yvanoers:
The job name was not checked for being empty. Fixes #656.
Also updated comments in grpc.go.

### DIFF
--- a/dkron/api_test.go
+++ b/dkron/api_test.go
@@ -159,6 +159,39 @@ func TestAPIJobCreateUpdateValidationValidName(t *testing.T) {
 	assert.Equal(t, http.StatusCreated, resp.StatusCode)
 }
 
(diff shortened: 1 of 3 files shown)


yvanoers:
Closing & reopening to rerun tests

Issue #661, opened 2019-12-13
Differing date formats in Executions and Webhooks


davidgengenbach:
**Describe the bug**
The date format differs between in
- the job execution timestamps (= `started_at`, `finished_at`)
  - `2019-12-12T22:57:47.782Z`
    - https://en.wikipedia.org/wiki/ISO_8601 ?
- the webhook template
  - `2019-12-13 01:03:00.024564011 +0000 UTC`

**To Reproduce**
1. Configure a webhook (with template)
2. Create a job
3. Wait for the execution of the job

**Expected behavior**
The execution timestamps should have the same format as the webhook notifications.

**Specifications:**
 - OS: `Linux david 5.3.14-200.fc30.x86_64 #1 SMP Mon Dec 2 15:57:50 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
 - Version `Dkron 2.0.0`

**Additional context**

From the [golang time documentation](https://golang.org/pkg/time/#Time.String): 
>  The returned string is meant for debugging; for a stable serialized representation, use t.MarshalText, t.MarshalBinary, or t.Format with an explicit format string. 

The date format in the job executions are the "right" ones - I would think. 

I would propose using the `t.MarshalText` as suggested by the golang people!

These are the places where the fields are defined/marshalled:
Webhooks: https://github.com/distribworks/dkron/blob/master/dkron/notifier.go#L86
Executions: https://github.com/distribworks/dkron/blob/master/dkron/execution.go#L18

yvanoers:
@davidgengenbach I suspect the format in the webhook template is how it is because it is intended to be human readable, whereas the API returns computer readable values. Maybe @victorcoder can confirm this.

Nevertheless it is better to use `t.MarshalText` or the likes, as you mentioned. All that needs to be agreed on is the right format, and I'll whip up a PR.

victorcoder:
Agree, better to use MarshalText, thanks for the finding @davidgengenbach

victorcoder:
Fixed in #665

PR #665, opened 2019-12-28
Use time type instead of a preformatted string for times in notification template

yvanoers:
Addresses #661.
The date format used is the same as time.String uses, with the exception of the `+-m=nnnn` value at the end. The m value is not always present, so this is reasonably backwards compatible, but not 100% guaranteed.

If the format should be changed, please let me know.

### DIFF
--- a/dkron/notifier.go
+++ b/dkron/notifier.go
@@ -9,6 +9,7 @@ import (
 	"net/textproto"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/jordan-wright/email"
 	"github.com/sirupsen/logrus"
@@ -74,8 +75,8 @@ func (n *Notifier) buildTemplate(templ string) *bytes.Buffer {
 		Report        string
 		JobName       string
 		ReportingNode string
-		StartTime     string
-		FinishedAt    string
+		StartTime     time.Time
+		FinishedAt    time.Time
 		Success       string
 		NodeName      string
 		Output        string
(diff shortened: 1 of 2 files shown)


davidgengenbach:
> Addresses #366 
> 
> If the format should be changed, please let me know.

I would **really** propose using the same format for both the Webhook and in other places! Having differing formats will only complicate things. The proposed format in this PR does not seem to be standardized?
Using the standard will greatly ease integration with other software.
For example, when using the webhook timestamp in ElasticSearch, I had to write a custom parser as ElasticSearch did not recognize the date automatically (same applies to Java).

As in https://github.com/distribworks/dkron/issues/661, I would propose using 
https://en.wikipedia.org/wiki/ISO_8601
It seems to be the reasonable date-format - it is also used for the execution timestamps, e.g. `started_at`.

> The date format used is the same as time.String uses, with the exception of the `+-m=nnnn` value at the end. The m value is not always present, so this is reasonably backwards compatible, but not 100% guaranteed.

Being backward-compatbility is nice, yes - but as you said, the compatibility with this solution only works partially, so erronous behaviour might only happen in some cases (making debugging the change more difficult). As far as I know, the webhook functionality is only documented in the CLI options and there is no dedicated "Usage" article about it?
I would make this a breaking change, i.e. changing the date format to the standard.

yvanoers:
I've changed it so the time itself goes into the template.
This way, the template can specify how to use the time, like so:
`{{.StartTime.Format "your format here"}}`

If you want to use ISO8601/RFC3339, this will allow you to do so, or whatever format anyone would want.

Bonus: this is fully backwards compatible. Simply using `{{.StartTime}}`, will default to the String() representation.

I was going for more flexibility, not backwards compatibility, but it's nice to have both.

PR #684, opened 2020-02-18
WIP: trim plugin size

fopina:
This is WIP but a retry for #537 

Now it even makes more sense as plugin should remain the same old optimum size (around 11M) while current size is much bigger than before (due to all the new dependencies in dkron v2).

Need to tweak a few things yet, but is this something of interest? Any quick comments on the structure/re-packaging, as in, moved `proto` generated .go files to `plugin/types` plus the required definitions to have all the plugin dependencies isolated within `plugin`


victorcoder:
Yes @fopina I'm interested on this, thanks!

PR #688, opened 2020-03-01
change disk storage to  mysql

roomcat:
remove raft FSM and add mysql storage to store job and execution data


victorcoder:
Hi @roomcat just to let you know that this kind of effort is not in line with the philosophy of the project. Introducing mysql as storage imply a non HA component in the architecture, and we aim to be fully HA.

Thanks

PR #689, opened 2020-03-05
Fixes crash when plugin configuration is not defined in a job

andreygolev:
If there's no plugin configuration in a job definition, dkron will crash with error:
panic: assignment to entry in nil map

This PR fixes it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,23 +1,22 @@
 name: Test
-on: [pull_request]
+on: [push,pull_request]
 jobs:
   build:
     name: Test
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.13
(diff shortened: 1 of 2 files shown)


victorcoder:
I don't know why this test is failing

victorcoder:
@andreygolev can you modify .github actions to this?

```
name: Test
on: [push,pull_request]
jobs:
  build:
    name: Test
    runs-on: ubuntu-latest
    steps:

    - name: Set up Go 1.14
      uses: actions/setup-go@v1
      with:
        go-version: 1.14
      id: go

    - name: Check out code into the Go module directory
      uses: actions/checkout@v2

    - name: Get dependencies
      run: go get -v -t -d ./...

    - name: Test without docker
      run: go test -v -timeout 200s ./...
```

PR #690, opened 2020-03-05
Graceful shutdown

andreygolev:
When I started to test graceful shutdown of worker type agents, it appeared that they don't wait for their running jobs before shutdown.
If I recall correctly, plugin.CleanupClients() kill plugins immediately, causing childs (in case of shell executor) to die.

This PR does the following:
- SIGTERM received
- Agent leaves cluster to not accept new jobs 
- Agent waits for active jobs to finish during 3h
- Exits when running jobs finished or gracefulTimeout reached

Note: Running jobs successfully can send ExecutionDone to leader even after agent left cluster.

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/distribworks/dkron/v2/dkron"
-        "github.com/hashicorp/go-plugin"
+	"github.com/hashicorp/go-plugin"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -130,7 +130,7 @@ WAIT:
 		time.Sleep(1 * time.Second)
 	}
 
-        plugin.CleanupClients()
+	plugin.CleanupClients()
 
 	close(gracefulCh)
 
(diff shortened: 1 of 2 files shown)


cmd/agent.go
```diff
@@ -110,18 +109,28 @@ WAIT:
 	}
 
 	// Attempt a graceful leave
-	gracefulCh := make(chan struct{})
 	log.Info("agent: Gracefully shutting down agent...")
 	go func() {
-		plugin.CleanupClients()
```
victorcoder: When is this going to happen?

andreygolev: I've tested it only with containers. Container exits, and the everything that was running inside dies also. So it doesn't matter in my case :)
But you're right, not everyone is using containers.
I'll add plugin.CleanupClients() somewhere below

cmd/agent.go
```diff
@@ -110,18 +110,30 @@ WAIT:
...
+	for {
+		log.Info("Waiting for jobs to finish...")
+		if agent.GetRunningJobs() < 1 {
+			log.Info("No jobs left. Exiting.")
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+        plugin.CleanupClients()
```
victorcoder: Is this gofmt'ed?

andreygolev:
Sorry, edited it with vi first :)

cmd/agent.go
andreygolev: Sorry, edited it with vi first :)

PR #692, opened 2020-03-05
Execution load balancing

andreygolev:
serf.Members() seems to usually return same order of nodes.

When we have job definition with tag "label:1", the job is executed usually on the same worker agent .
This PR adds shuffling, before returning a node for execution.

After shuffling is added, we see good load distribution among worker pool with matching tag.

Screenshot 2020-03-05 at 13 44 06

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -780,8 +780,9 @@ func (a *Agent) processFilteredNodes(job *Job) ([]string, map[string]string, err
 					break
 				}
 				nodes = append(nodes, candidates[0])
-				candidates = candidates[:1]
+				candidates = candidates[1:]
 			}
+			candidates = nil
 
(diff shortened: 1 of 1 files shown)


andreygolev:
tests didn't work by unknown reason

victorcoder:
It should work if you rebase master here

andreygolev:
> It should work if you rebase master here

Rebase didn't help

victorcoder:
Check the tests, it's failing in `Test_processFilteredNodes`

andreygolev:
There was a bug, indeed. Very bad one :) Now fixed.

victorcoder:
> There was a bug, indeed. Very bad one :) Now fixed.

Good to have tests then 👍

PR #693, opened 2020-03-05
crash fix during server failover

andreygolev:
This is not the best implementation, but fixes crash when worker agent after job execution tries to send ExecutionDone to a server in a cluster, but server is already dead.

Don't know how to do it better than net.Dial() for now.
As an addition it would be imho better to add retries with different members


dkron/invoke.go
```diff
@@ -61,7 +62,7 @@ func (a *Agent) invokeJob(job *Job, execution *Execution) error {
 	execution.Success = success
 	execution.Output = output.Bytes()
 
-	rpcServer, err := a.selectServerByKey(execution.Key())
```
victorcoder: can `selectServerByKey` be removed?

andreygolev: according to the code, there are no occurrences.

victorcoder:
@andreygolev did you experienced crashes with the old method? I don't get the point of doing a net.Dial check, the `LocalServers` method already returns the servers available that are live. The timming is the same as in this proposal.

andreygolev:
Yes, I've experienced crashes with the old method.
It was clearly visible in debug logs that worker agent was trying to send RPC call to an IP of an already dead server.
LocalServer can return a node that in reality is already dead, but fact still not yet confirmed by cluster, because failover takes some time.

net.Dial just checks if you can connect to the IP:port. If not, then try next one.

andreygolev:
Better approach would be to add retries imho

dkron/invoke.go
```diff
@@ -72,18 +72,21 @@ func (a *Agent) invokeJob(job *Job, execution *Execution) error {
...
-		ch.Add(p.RPCAddr.String())
+		peers = append(peers, p.RPCAddr.String())
 	}
 
-	peerAddress := ch.Get(key)
-	if peerAddress == "" {
-		return "", ErrNoSuitableServer
+	for _, peer := range peers {
+		log.Debugf("Checking peer: %v", peer)
+		conn, err := net.DialTimeout("tcp", peer, 5*time.Second)
```
victorcoder: 5 seconds for timeout seems too much

andreygolev: Safe choice :)
What about 1 second?

victorcoder: It should work well

andreygolev: here we go

PR #702, opened 2020-03-11
BuntDB with delete executions and fixed tests

andreygolev:
This is a PR based on buntdb branch.
Differences:
* Added deleting of executions in DeleteJob method.
* Added ignore self in listing in listTxFunc. (wasn't in buntdb branch)
* Fixed check for metadata in job in DeleteJob method. (tests were failing)
* Some renames
* BadgerDB fully removed
* Snapshot and Restore methods in Store returning nil.  Looks like these are required for Raft, but we don't need to snapshot or restore BuntDB as it's restored via Raft.


dkron/store.go
```diff
@@ -32,9 +27,10 @@ var (
 
 // Store is the local implementation of the Storage interface.
 // It gives dkron the ability to manipulate its embedded storage
-// BadgerDB.
+// BuntDB.
 type Store struct {
-	db     *badger.DB
+	agent  *Agent
```
victorcoder: You can remove this as it's not used

dkron/store.go
```diff
@@ -44,89 +40,45 @@ type JobOptions struct {
...
+		if _, _, err := tx.Set(jobKey, string(jb), nil); err != nil {
 			return err
 		}
 
 		return nil
 	}
 }
 
-// DB is the getter for the BadgerDB instance
-func (s *Store) DB() *badger.DB {
```
victorcoder: This is needed for Pro version

dkron/store.go
```diff
@@ -627,27 +545,38 @@ func (s *Store) SetExecution(execution *Execution) (string, error) {
...
 }
 
 // Shutdown close the KV store
 func (s *Store) Shutdown() error {
 	return s.db.Close()
 }
 
-// Snapshot creates a backup of the data stored in Badger
+// Snapshot stub for Raft
 func (s *Store) Snapshot(w io.WriteCloser) error {
```
victorcoder: You can also remove this as we don't use this as Raft store

dkron/store.go
```diff
@@ -627,27 +545,38 @@ func (s *Store) SetExecution(execution *Execution) (string, error) {
...
-	_, err := s.db.Backup(w, 0)
-	return err
+	return nil
 }
 
-// Restore load data created with backup in to Badger
-// Default value for maxPendingWrites is 256, to minimise memory usage
-// and overall finish time.
+// Restore stub for Raft
 func (s *Store) Restore(r io.ReadCloser) error {
```
victorcoder: You can also remove this as we don't use this as Raft store

andreygolev: If I remove this method (same for Snapshot), then other things claim about Restore/Snapshot methods.

```
diff --git a/dkron/fsm.go b/dkron/fsm.go
index 03474bf3..69505961 100644
--- a/dkron/fsm.go
+++ b/dkron/fsm.go
@@ -139,7 +139,7 @@ func (d *dkronFSM) Snapshot() (raft.FSMSnapshot, error) {
 // Restore stores the key-value store to a previous state.
 func (d *dkronFSM) Restore(r io.ReadCloser) error {
        defer r.Close()
-       return d.store.Restore(r)
+       return nil
 }
 
 type dkronSnapshot struct {
@@ -147,12 +147,6 @@ type dkronSnapshot struct {
 }
 
 func (d *dkronSnapshot) Persist(sink raft.SnapshotSink) error {
-       if err := d.store.Snapshot(sink); err != nil {
-               sink.Cancel()
-               return err
-       }
-
-       // Close the sink.
        if err := sink.Close(); err != nil {
                return err
        }
```

Should I return nil in place like above instead?
Because, if I remove Snapshot and Persist in fsm.go, then raft.NewRaft() and raft.RecoverCluster() claim that they're missing methods Restore and Snapshot:
```
cannot use fsm (variable of type *dkronFSM) as raft.FSM value in argument to raft.NewRaft: missing method Restore
cannot use tmpFsm (variable of type *dkronFSM) as raft.FSM value in argument to raft.RecoverCluster: missing method Restore
```

victorcoder: Oh yes, because it's defined in the Storage interface which FSM uses. Restore and Snapshot should be implemented with https://godoc.org/github.com/tidwall/buntdb#DB.Load and https://godoc.org/github.com/tidwall/buntdb#DB.Save

andreygolev: I wonder what's the point having Save and Load if BuntDB is restored through raft which is persisted and even replicated from other nodes?

victorcoder: This is in case of server restart, the point is that Raft log works with a combination of snapshots and log, when you restart a server that has been compacted, it will load the snapshot and apply the logs since the last snapshot. If snapshots are not saved, the storage won't be in sync.

andreygolev: Alright, gotcha. My confusion was caused, because we run Dkron in Kubernetes without any state persistence. Each pod restart causes full wipe both of Raft and Store.
We just rely on HA and replication there. In case everything is really bad, we just restore job definitions with Terraform.

dkron/store.go
```diff
@@ -44,86 +39,46 @@ type JobOptions struct {
...
-		WithValueLogLoadingMode(options.FileIO).
-		WithLogger(log)
+type kv struct {
+	Key   string
+	Value []byte
+}
 
-	db, err := badger.Open(opts)
+// NewStore creates a new Storage instance.
+func NewStore(a *Agent) (*Store, error) {
```
victorcoder: Remove agent param

PR #704, opened 2020-03-12
Kubernetes support

andreygolev:
This PR contains:
* Kubernetes manifests for running Dkron server in HA mode
* Kubernetes manifests for running Dkron worker agents that are able to discover and join Dkron server
* Building in Docker and baking image in one step.
   This will allows to build own image on any platform
* Compiling with CGO_ENABLED=0.  
   Dkron didn't work on older Linux distros without this if binary was compiled on some newer distro.
   In our case we were just copying precompiled binary to our existing apps to use dkron with shell executor plugin.
* Added jq to Docker image, as it's required for Kubernetes, to parse Dkron API output.

Kubernetes has a specifics that IP addresses of pods are changed on each restart. This is not working good with Dkron.
So, Dkron is launched with wrapper script that is doing all the job of removing peers from cluster

This solution comes with no storage persistence, as it's currently working bad with Raft in Kubernetes.
In our case we store job definitions in Terraform, and this is our source of truth and a backup.
In case we lose cluster, we can just bootstrap a new one and apply Terraform in seconds, which will configure Dkron through API

### DIFF
--- a/Dockerfile.hub
+++ b/Dockerfile.hub
@@ -1,19 +1,31 @@
-FROM alpine
+FROM golang:1.14-alpine as builder
 LABEL maintainer="Victor Castell <victor@victorcastell.com>"
 
+RUN mkdir -p /app
+WORKDIR /app
+
+ENV GO111MODULE=on
+COPY go.mod go.mod
+COPY go.sum go.sum
+RUN go mod download
+RUN apk add build-base
+
+COPY . .
+RUN CGO_ENABLED=0 go install ./...
+
+FROM alpine
+# JQ Required for prestop hook to remove peer from cluster on pod delete action
 RUN set -x \
-	&& buildDeps='bash ca-certificates openssl tzdata' \
-	&& apk add --update $buildDeps \
-	&& rm -rf /var/cache/apk/* \
-	&& mkdir -p /opt/local/dkron
+        && buildDeps='bash ca-certificates openssl tzdata jq' \
+        && apk add --update $buildDeps \
+        && rm -rf /var/cache/apk/* \
+        && mkdir -p /opt/local/dkron
+
+COPY --chown=0:0 --from=builder /go/bin/dkron* /opt/local/dkron/
 
 EXPOSE 8080 8946
 
 ENV SHELL /bin/bash
 WORKDIR /opt/local/dkron
-
-COPY dkron .
-COPY dkron-* ./
 ENTRYPOINT ["/opt/local/dkron/dkron"]
-
 CMD ["--help"]
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -413,6 +413,10 @@ func (a *Agent) setupSerf() (*serf.Serf, error) {
 	serfConfig.UserCoalescePeriod = 3 * time.Second
 	serfConfig.UserQuiescentPeriod = time.Second
 
+	if a.config.Kubernetes {
+		serfConfig.ReconnectTimeout = 5 * time.Second
+	}
+
 	// Create a channel to listen for events from Serf
 	a.eventCh = make(chan serf.Event, 64)
 	serfConfig.EventCh = a.eventCh
--- a/dkron/config.go
+++ b/dkron/config.go
@@ -153,6 +153,10 @@ type Config struct {
 
 	// StatsdAddr is the statsd standard server to be used for sending metrics.
 	StatsdAddr string `mapstructure:"statsd-addr"`
+
+	// Kubernetes is used to set faster reap timeout, as node won't come back with same
+	// IP after pod restart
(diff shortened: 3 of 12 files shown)


victorcoder:
Personally, I don't have a way to test all this code as I don't use K8's, I prefer this specific config should be stored in its own repo.

Regarding the reap time, it can be set to be configurable with no reference to Kubernetes.

drauschenbach:
@andreygolev great work! I've repackaged the assets in this merge request as a Helm chart at https://github.com/Megalithic-LLC/dkron-helm.

@victorcoder I agree and so far I see no need for Go source code modifications or Docker image changes just to support Kubernetes.

jjsaunier:
I don't really get the whole complexity arround retry / joining of this chart. 

adding this annotation to leader / worker

```
    annotations:
        consul.hashicorp.com/auto-join-port: "8946"
```

See https://www.consul.io/docs/install/cloud-auto-join#kubernetes-k8s. Plus, there is no reason to dkron to be able to patch other pods, sounds wrong to me

Increasing retry / reconnect timeout may help when new node join / first init (pulling images etc) takes more time by adding those flags on dkron:

```
 - "--serf-reconnect-timeout=5s"
-  "--retry-interval=10s"
```

Open question here, I'm using dkron as DaemonSet, i'm not sure about the benefit to use it as sts, moreover without affinity / antifinity placements 

As bonus, if you use prometheus : 

```
{{- if .Values.monitoring.enabled }}
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
    name: "dkron-prometheus-exporter"
    namespace: {{ .Release.Namespace }}
    labels:
        name: "dkron-prometheus"
spec:
    endpoints:
        -   port: http
            path: /metrics
            interval: 10s
    jobLabel: "dkron-prometheus-exporter"
    namespaceSelector:
        matchNames:
            - {{ .Release.Namespace }}
    selector:
        matchLabels:
            app: dkron
{{- end }}
```

andreygolev:
Hey!

Regarding complexity and these scripts. It was all taken from Consul helm chart back in the days.
In reality, this is true, this all is unneeded now and works without that just fine, with simply putting something like 
`retry-join: ["provider=k8s label_selector=` to a config.

I can agree that STS is not a necessary thing, but
STS there is just for less harmful upgrades of scheduler (server), where we can upgrade leader node last, manually.
You can't tell Kubernetes in what order to restart pods.
STS gives us something like controlled pod restarts during upgrades, to not trigger failovers, for example.

Dkron didn't really like rapid failovers back in the days and easily could end up with split brain, when some members don't know about each other after you simply do a rolling upgrade.
That's why STS all there for "server/scheduler" part

In my case, I have strictly dedicated 3 pods for "server" part and the agents are joining to server from different namespaces and projects. 

WRT pod labels sidecar.
If we don't remove a peer from a Raft before restarting pod, Dkron may end up with split brain that was mentioned above.
You can tell Dkron to remove a raft peer only by telling this to a leader node.
That's why we need to know who's leader.
Maybe in case where hostname of a pod (like with ds or deployments) is always different this won't be a problem at all, and dead ex-peers will be reaped from Raft some day.

drauschenbach:
The `provider=k8s` functionality is specifically documented to explain that it can be used to join external agents to an EXISTING kubernetes-based Dkron cluster.

I have tried to use it beyond its documented capabilities, and tried to launch a 3-node StatefulSet cluster with it as you describe above, and indeed I ended up with 3 split-brained nodes that each thought they were the master.

jjsaunier:
I will try to reproduce the split brain issue, but i'm pretty sure it's coming from identity of worker / leader. In STS hostname remain the same and that's the whole point, dkron identify them with hostname. With my current model, if worker / leader is down / recreated or added it, it have a new identity anyway so I don't have to evict explictly the affected member to add the member. The unreachable member have it's own lifecycle managed by dkron while k8s already know the pod is down and reschedule a new instance (with it's own identity) and can join the dkron cluster without identity collision / wait that dkron evict it to add the new one (and avoid falling in timeout, where if a leader can't join, it will create it's own cluster)

That's why i'm not fan of STS for dkron, we should not care to identify member and ensure they are in the pool, it's a cattle model. I hear the argument to be able to prioritize upgrade of workload - I'm agree we should be able to upgrade leader first but the downside is too much. Daemonset and Deployment are a good fit IMO regarding the topology of cluster

nikunj-badjatya:
@andreygolev 
We are also trying to setup dkron on K8S as a STS. 
Can you share your experience of running dkron on K8S ? there seem to be very less documentation on this.
I have posted some questions here https://github.com/distribworks/dkron/issues/1191 would you be able to help ? Thanks

PR #705, opened 2020-03-13
Don't block execution if concurrency is allowed

andreygolev:
Maybe I get it wrong, but it looks to me that even if concurrency set to allow, isRunnable() still returns false


victorcoder:
You are wrong, and the test says it so, maybe the variable name is not the most intuitive one but, as the log message say, that check disallow the execution while the last execution is still being broadcasted to the targets, this is not allowed because if there is a high frequency of executions < broadcast time, serf queue will grow without limit making all other messages stay in queue and finally collapsing the system.

Hope is makes sense.

PR #719, opened 2020-04-06
Streaming executions part 2

victorcoder:
Implement realtime output streaming for executors.

Kudos to @fopina for providing the work on streaming plugins in #622

closes #622


fopina:
That was fast!!

PR #739, opened 2020-04-27
Add job name tags to log events to improve debugging

sysadmind:
Some of the log calls had the job_name tag, but others did not. This adds them to those calls that were missing to improve log correlation, especially when there are many jobs all with the same schedule (i.e. @every 1m) and these execution log entries are all intertwined.

I kept the "job_name" key to match other logging calls in this file, but it seems to be inconsistent with other files (store.go uses "job")


sysadmind:
Quite happy to convert while I'm already making changes. All occurrences now use `job`

Issue #494, opened 2019-02-18
Expose metrics in prometheus format


sc0rp10:
Hi! I am using dkron for many tasks in my cluster - this is a great, simple and powerful software, thanks for it!
I have couple of feature requests that can be useful for another users.

It would be nice, if Dkron will expose metrics in prometheus format by some url like `/metrics` for advanced analytics/alerting/vizualizing jobs and their statuses.
I know about statsd + statsd_exporter, but IMO now prometheus is a industry standard and will be great if we can combine together dkron and prometheus without any another software.

victorcoder:
Hi @sc0rp10, thanks for the proposal, I was thinking in improving the metrics method in Dkron and I have Prometheus on top of the list.

Expect some changes in the near future.

victorcoder:
Done in #740

PR #740, opened 2020-04-28
Add prometheus endpoint for metrics

sysadmind:
The existing metrics package used in dkron already supports prometheus, so this just adds in the necessary pieces to hook it up to the API. There is a config value to enable serving this endpoint to allow users the choice with a default of false to keep today's functionality. I know that it was discussed in #494 that the metrics system was likely going to be updated, but this seems like an easy step in the meantime.

The documentation for metrics was updated to align with current configuration keys.


sysadmind:
I changed to using prometheus.DefaultRegistry. I initially thought that the metrics package was already setting up the metrics that come along with the DefaultRegistry, but when I went to compare more closely after your comment, I realized that the DefaultRegistry actually has a lot more metrics and they are more standardized across the prometheus community than those configured by the metrics package so I changed the PR.

PR #745, opened 2020-05-04
Retries

andreygolev:
This is not the best code, but it works so far :)
Just want to show my approach for this problem.

I'm hardcoding serf query timeout for 10 seconds, because default quert timeouts differs, based on some other values in Serf. 10 seconds imho is enough for Dkron purposes to send query and get back an ack, as response payload is not used at all for now. Also we have retries :)

When sending a Serf query, I'm also sending expected ack response time with payload. If agent is finally receiving a serf query later than timeout, it shouldn't execute a job. Without this safety check, agents will execute everything that had retry attempts, resulting in multiple executions of a job, which is not desired.

Also, in job.go I've added random delay, because problem with missing jobs now clearly visible that appears only when we have lots of simultaneous schedules which generates lots of Serf messages (queue still fine during this). So, I'm a bit delaying to reduce "load" on Serf. Just as an additional safety measure :) Maybe it's fine not have it.

Retries do happen in queries.go.
There's just hardcoded 10 attempts, with increasing sleep between retries. Poor man's backoff :)
Though on our quite high load, I haven't seen more than 3 retry attempts for a job. Retry amount really changes on query timeout value. First I had 2 seconds timeout, and it was quite bad. Also I've added logging which shows the time diff when the node finally responded, and numbers sometimes are quite high.

Also here we're not waiting for Serf query timeout if all expected nodes have responded quicker than timeout. And after that we also checking if we have acks only from expected nodes.

Kinda this.
This is not a PR, just an idea :)

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -410,6 +410,7 @@ func (a *Agent) setupSerf() (*serf.Serf, error) {
 	serfConfig.QuiescentPeriod = time.Second
 	serfConfig.UserCoalescePeriod = 3 * time.Second
 	serfConfig.UserQuiescentPeriod = time.Second
+	serfConfig.ReconnectTimeout = time.Duration(config.SerfReconnectTimeout) * time.Second
 
 	// Create a channel to listen for events from Serf
 	a.eventCh = make(chan serf.Event, 2048)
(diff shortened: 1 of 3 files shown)


victorcoder:
@yvanoers, a refactor for job executions is merged on master as you know tests are looking good, thanks for the effort here.

PR #772, opened 2020-05-25
busyhandler execution sorting

fopina:
Currently, busyHandler does no sorting, and the web UI doesn't either.
The result is a page where it is hard to follow any output as executions keep switching order:
Peek 2020-05-25 00-11

Sorting can be done both in the UI or in the busyHandler method.
In the UI it probably saves that (small yet existing) processing from dkron server, but then it shouldn't return an array.  
As an API consumer I expect a JSON array response to have some sort of order (usually, from a DB, it would be the cheapest order, the insertion order).
Regardless of the sorting criteria, the same request with the same set of results should always have the same order.
So I'd say it could make sense to leave the UI to order the response but turn it into a mapping/hash (`execution_name: execution`) or simply sort in `busyHandler` with some common criteria such as `jobName` or `startTime`.

I chose the latter in this PR, let me know if you'd rather have another approach.

End result:

Peek 2020-05-25 01-49

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sort"
 
 	"github.com/gin-contrib/expvar"
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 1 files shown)


fopina:
probably member list (in the dashboard page) could use a sort as well ?

victorcoder:
> probably member list (in the dashboard page) could use a sort as well ?

It already does but in the UI https://github.com/distribworks/dkron/blob/master/templates/index.html.tmpl#L25

Issue #794, opened 2020-06-29
Invalid memory address or nil pointer dereference


clarifysky:
**Describe the bug**
I have three nodes, after 5 days of running, one of them crashed.
Actually, even one node, after 5 days of running, it may also crash.

**To Reproduce**
1. Launch dkron: `nohup dkron agent >> /var/log/dkron.log &`.
2. Create several jobs (per second, per miniute, per hour, per day, many kinds).
3. Let it keep running for 5 days or more. 
4. After 5 days of running, watch the process of dkron (it may have crashed), then view the log.

**Expected behavior**
It can run steadily at least for about half of a month.

**Screenshots**

** Specifications:**
 - OS: linux
 - Version: 3.10.0

**Additional context**
Below is the crash log:

```log
time="2020-06-25T05:19:12+00:00" level=debug msg="dkron: shutting down leader loop" node=abc
time="2020-06-25T05:19:12+00:00" level=debug msg="scheduler: Stopping scheduler" node=abc
time="2020-06-25T05:19:12+00:00" level=info msg="2020-06-25T05:19:12.758+0000 [INFO]  raft: aborting pipeline replication: peer=\"{Voter abc01 192.168.10.1:6868}\""
time="2020-06-25T05:19:12+00:00" level=info msg="dkron: cluster leadership lost" node=abc
time="2020-06-25T05:19:12+00:00" level=info msg="dkron: monitoring leadership" node=abc
time="2020-06-25T05:19:12+00:00" level=info msg="2020-06-25T05:19:12.826+0000 [ERROR] raft: failed to appendEntries to: peer=\"{Voter abc02 192.168.10.2:6868}\" error=\"dial tcp 192.168.10.2:6868: i/o timeout\""
time="2020-06-25T05:19:14+00:00" level=info msg="2020-06-25T05:19:14.146+0000 [DEBUG] raft: lost leadership because received a requestVote with a newer term"
time="2020-06-25T05:19:14+00:00" level=info msg="2020-06-25T05:19:14.223+0000 [ERROR] raft: failed to heartbeat to: peer=192.168.10.2:6868 error=\"dial tcp 192.168.10.2:6868: i/o timeout\""
time="2020-06-25T05:19:17+00:00" level=debug msg="job: Run job" job=abc01_check_tomcat node=abc schedule="@every 1m"
time="2020-06-25T05:19:17+00:00" level=debug msg="store: Retrieved job from datastore" job=abc01_check_tomcat node=abc
[...]

PR #799, opened 2020-07-09
Fix: free s.Cron on proper opportunity to avoid unexpected crash.

clarifysky:
Ignore the running jobs and make `s.Cron` to nil may cause whole process crashed.

After digging, I found that `s.Cron` will be `nil` when `revokeLeadership()` be called, but you will never know when `revokeLeaderShip()` will be call. 

Maybe when calling `revokeLeaderShip()`, there is also one job is still running (they are in different goroutines, they are both running), it needs `s.Cron`, but before it utilize `s.Cron`, `s.Cron` has been freed by `revokeLeadership()`, then the whole prcess will crash.

Below is the crash point, `s.Cron.Entries()`:

```go
// GetEntry returns a scheduler entry from a snapshot in
// the current time, and whether or not the entry was found.
func (s *Scheduler) GetEntry(jobName string) (cron.Entry, bool) {
	for _, e := range s.Cron.Entries() {
		j, _ := e.Job.(*Job)
		if j.Name == jobName {
			return e, true
		}
	}
	return cron.Entry{}, false
}
```

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -240,6 +240,7 @@ func (a *Agent) Stop() error {
 
 	if a.config.Server && a.sched.Started {
 		a.sched.Stop()
+		a.sched.ClearCron()
 	}
 
 	if err := a.serf.Leave(); err != nil {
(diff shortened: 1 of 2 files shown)


clarifysky:
This should fix the issue: #794 

After apply these code, I found that My two nodes keep running for at least 9 days steadily, but before, they will crash after 5 days.

victorcoder (approved):
LGTM as this looks quite un-harmful

PR #850, opened 2020-11-11
Update release.yml

fopina:
I was just about to open a PR to add fetch-depth after seeing the release changelog today
I used it the first time myself when master was v1 (and depth was not 1 by default)

Probably worth pinning checkout to `v2` to avoid future breaking changes like this (as suggested in the [issue](https://github.com/goreleaser/goreleaser-action/issues/56))


victorcoder (approved):
Yup, thanks!

PR #864, opened 2020-12-10
React admin Web UI

victorcoder:
Effort to rewrite the UI using https://marmelab.com/react-admin/

Just missing the Dashboard graph, no really useful TBH. Other features work better that before, also it's now much easier to implement improvements to the search, etc.

cc/ @yvanoers


fopina:
<3
Now I can't wait to find the time to try to rewrite to preact for smaller browser footprint!:P

PR #842, opened 2020-11-02
Fix multitag cardinality bug

yvanoers:
Fixes #821.

PR #865, opened 2020-12-10
Fix uneven balancing

yvanoers:
Fixes uneven balancing, reported in #842 .
The actual fix is in agent.go line 717.
Also renamed `nameIndex` to `names`, it is better suited to what it holds.
Amended a test to verify the balancing. This test might result a false negative, because it is testing randomness.


victorcoder (approved):
LGTM, thanks!

PR #878, opened 2020-12-23
Added some query filters on Job executions

MGSousa:


victorcoder (changes requested):
LGTM but could you rename `JobExecutionsOptions` to `ExecutionOptions`?

victorcoder (changes requested):
This won't work unless executions are stored in json in BuntDB, now they are stored in binary protobuf, or am I missing something?

MGSousa:
Yes, the executions are stored in binary Protobuf with `proto.Marshal` but these changes will be ready for JSON marshaling in the future (if migrated), also it contains an option to order all executions by ASC or DESC and by itself is helpful.

MGSousa:
> LGTM but could you rename `JobExecutionsOptions` to `ExecutionOptions`?

Done!

dkron/storage.go
```diff
@@ -16,7 +16,7 @@ type Storage interface {
 	SetExecutionDone(execution *Execution) (bool, error)
 	GetJobs(options *JobOptions) ([]*Job, error)
 	GetJob(name string, options *JobOptions) (*Job, error)
-	GetExecutions(jobName string, timezone *time.Location) ([]*Execution, error)
+	GetExecutions(jobName string, timezone *time.Location, opts *ExecutionOptions) ([]*Execution, error)
```
victorcoder, line 19: It would be great to refactor `timezone` into `ExecutionOptions`

victorcoder (changes requested):

victorcoder:
> Yes, the executions are stored in binary Protobuf with proto.Marshal but these changes will be ready for JSON marshaling in the future (if migrated), 

Then why implement this if it does not work? You can implement json marshal/unmarshal with fallback to proto.Unmarshal as I did with jobs, take a look at https://github.com/distribworks/dkron/blob/master/dkron/store.go#L309-L315

> also it contains an option to order all executions by ASC or DESC and by itself is helpful.

You mean ASC or DESC by the key?

MGSousa:
> You mean ASC or DESC by the key?

Indeed.

dkron/storage.go
MGSousa: Done

victorcoder (approved):
LGTM

PR #879, opened 2021-01-01
Use correct wrapping and monospace font for execution's output

sc0rp10:
Hi,
I like the new UI and using it everyday, but I think we missed something important - correct command output.
For example, I've created job that shows recursively directory structure.

In current stable version it looks not very useful:

I've applied some CSS changes to fix it:

I've added `max-height` too to prevent problems with huge outputs, but I'm not sure that I used right value for such rule.

@victorcoder what do you think about these changes?


victorcoder (approved):
LGTM thanks!

PR #886, opened 2021-01-08
simpler goreleaser.yaml

fopina:
not a mindblowing PR but this simpler goreleaser.yaml with anchors makes it easier to comment out archs for local builds :)


victorcoder (approved):
LGTM

PR #890, opened 2021-01-14
Confusion in nth fields in documentation

denrou:
First field in cron expression represents seconds and not minutes


victorcoder (approved):
LGTM

PR #897, opened 2021-01-25
Provide filter query by job displayName, add pristine jobs reports

MGSousa:
- Provide filter query by job displayName in main search
- Change var DKRON_FAILED_JOBS to receive only "failed jobs"
- Added pristine jobs reports (jobs that have not yet been performed)


victorcoder (changes requested):
LGTM but I miss the change in React UI to use PRISTINE_JOBS

MGSousa:
> LGTM but I miss the change in React UI to use PRISTINE_JOBS

Done

dkron/ui.go
```diff
@@ -40,14 +40,17 @@ func (h *HTTPTransport) UI(r *gin.RouterGroup) {
...
+				successfulJobs, failedJobs, pristineJobs int
+			)
 			for _, j := range jobs {
 				if j.Status == "success" {
 					successfulJobs++
-				} else {
+				} else if j.Status == "failed" {
 					failedJobs++
+				} else {
+					pristineJobs++
```
yvanoers, line 53: The term "pristine" is confusing to me. To me it means 'unscathed', 'not worn by use'. I had to look through the code to see what the intent is, to discover that it denominates untriggered jobs.
IMHO using 'untriggered' is a lot more concise.

yvanoers (changes requested):

dkron/ui.go
MGSousa: Personally, I have used to declare something untouched (without execution/status changes ) but to be more clear, I will change to "untriggered".

dkron/store.go
```diff
@@ -125,6 +125,9 @@ func (s *Store) SetJob(job *Job, copyDependentJobs bool) error {
 	}
 
 	err := s.db.Update(func(tx *buntdb.Tx) error {
+		// set pristine as default for newly created jobs
+		job.Status = "pristine"
```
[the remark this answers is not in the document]
MGSousa: Yes, the job status gets updated according to execution status, but what is the problem with the default value getting lost? AFAIK the default value only gets used when a new job was created.

And secondly, I did it this way to allow by default **all** jobs (success/failed/untriggered) to continue showing without specifying filters.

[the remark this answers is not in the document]
yvanoers: The scenario I mean is when there is a job that is untriggered, and then dkron gets updated to the version with this change. That same untriggered job will still have status "" instead of the expected "pristine"/"untriggered".
I've tried to create this situation using this change and I found that the replay of the raft log is making it hard to reproduce. I need a bit more time to get a working example to prove (or disprove :) my point.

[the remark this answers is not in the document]
yvanoers: Here's the problem:
4 jobs, 1 successful, 1 failed, 1 untriggered (created before this change), 1 pristine (created after this change):
image

Filter these on Waiting to Run:
image

Only the one created after the change shows up, because the other one still has a Status of "".

This will occur when an untriggered job exists AND was persisted in a snapshot by raft before updating.

dkron/ui.go
```diff
@@ -58,11 +61,12 @@ func (h *HTTPTransport) UI(r *gin.RouterGroup) {
...
-				"DKRON_API_URL":         fmt.Sprintf("/%s", apiPathPrefix),
-				"DKRON_LEADER":          ln,
-				"DKRON_TOTAL_JOBS":      totalJobs,
-				"DKRON_FAILED_JOBS":     failedJobs,
-				"DKRON_SUCCESSFUL_JOBS": successfulJobs,
+				"DKRON_API_URL":         	fmt.Sprintf("/%s", apiPathPrefix),
+				"DKRON_LEADER":          	ln,
+				"DKRON_TOTAL_JOBS":      	totalJobs,
+				"DKRON_FAILED_JOBS":     	failedJobs,
+				"DKRON_PRISTINE_JOBS":   	pristineJobs,
```
yvanoers, line 68: ui/public/index.html also needs to be updated for the value to show up in the dashboard.
You don't have this file in your fork - pull from the master on distribworks/dkron to get that file.

yvanoers (changes requested):

dkron/ui.go
MGSousa: Done

dkron/store.go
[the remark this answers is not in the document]
MGSousa: Ok, I will do the required changes

dkron/ui.go
MGSousa: Done

victorcoder (changes requested):
whops! you will need to rebase your commits with master, or maybe open a new PR?

MGSousa:
@victorcoder Conflicts resolved

victorcoder:
@MGSousa but the history is polluted with commits from master, try to rebase and squash to one commit

MGSousa:
@victorcoder Commits already squashed, check if history yet prevails

yvanoers:
I would prefer it if 'pristine' was replaced with 'untriggered' in the front-end code as well, so it is consistent with the backend.
But I'll be ok with whatever @victorcoder wants in this regard.

Other than that, LGTM.

victorcoder (changes requested):
> I would prefer it if 'pristine' was replaced with 'untriggered' in the front-end code as well, so it is consistent with the backend.

Agree with @yvanoers suggestion

MGSousa:
> > I would prefer it if 'pristine' was replaced with 'untriggered' in the front-end code as well, so it is consistent with the backend.
> 
> Agree with @yvanoers suggestion

Done

ui/src/jobs/StatusFiled.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react";
```
victorcoder, line 1: The merge was not done correctly, this file is removed on master

dkron/ui.go
```diff
@@ -68,11 +71,12 @@ func (h *HTTPTransport) UI(r *gin.RouterGroup) {
...
-				"DKRON_API_URL":         fmt.Sprintf("/%s", apiPathPrefix),
-				"DKRON_LEADER":          ln,
-				"DKRON_TOTAL_JOBS":      totalJobs,
-				"DKRON_FAILED_JOBS":     failedJobs,
-				"DKRON_SUCCESSFUL_JOBS": successfulJobs,
+				"DKRON_API_URL":         	fmt.Sprintf("/%s", apiPathPrefix),
+				"DKRON_LEADER":          	ln,
+				"DKRON_TOTAL_JOBS":      	totalJobs,
+				"DKRON_FAILED_JOBS":     	failedJobs,
+				"DKRON_UNTRIGGERED_JOBS":   untriggeredJobs,
```
victorcoder, line 78: fmt

ui/public/index.html
```diff
@@ -47,4 +48,4 @@
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
   </body>
-</html>
+</html>
```
victorcoder, line 51:
```suggestion
</html>

```

ui/src/jobs/StatusField.tsx
```diff
@@ -1,10 +1,21 @@
...
-  return props.record === undefined ? null : (props.record[props.source] === 'success' ? <Tooltip title="Success"><SuccessIcon htmlColor="green" /></Tooltip> : <Tooltip title="Success"><FailedIcon htmlColor="red" /></Tooltip>);
+	if (props.record === undefined) {
+		return null
+	} else {
+		if (props.record[props.source] === 'success') {
+			return <Tooltip title="Success"><SuccessIcon htmlColor="green" /></Tooltip> 
+		} else if (props.record[props.source] === 'failed') {
+			return <Tooltip title="Error"><FailedIcon htmlColor="red" /></Tooltip>
+		} else {
+			return <Tooltip title="Waiting to Run"><PristineIcon htmlColor="blue" /></Tooltip>
```
victorcoder, line 16:
```suggestion
			return <Tooltip title="Waiting to Run"><UntriggeredIcon htmlColor="blue" /></Tooltip>
```

ui/src/jobs/StatusField.tsx
```diff
@@ -1,10 +1,21 @@
 import * as React from "react";
 import SuccessIcon from '@material-ui/icons/CheckCircle';
 import FailedIcon from '@material-ui/icons/Cancel';
+import PristineIcon from '@material-ui/icons/Timer';
```
victorcoder, line 4:
```suggestion
import UntriggeredIcon from '@material-ui/icons/Timer';
```

victorcoder (approved):
LGTM, thanks!

PR #907, opened 2021-02-23
real-time prometheus metrics from job executions

andreygolev:
This an initial PR for all metrics we get from shell job executions.

We do have more prometheus metrics that are extracted from finished jobs, but it's better to add them via another PR which depends on this PR.

* Job execution spent time
* Exit code
* Last execution timestamp (useful for alerting in case job is not successfully finished in expected time)


cobolbaby:
@vcastellm 

Is it possible to consider merging this feature.

vcastellm:
@andreygolev let us know if you want to work on this or we should take over, thanks.

cobolbaby:
@vcastellm Could you please take over?

vcastellm:
Taking over

PR #911, opened 2021-03-05
fix: don't kill job immediately if timeout is not defined

andreygolev:
Thanks @victorcoder for noticing this bad bug :)

https://github.com/distribworks/dkron/pull/906#discussion_r586845150


victorcoder (approved):
LGTM thx for fixing!

PR #923, opened 2021-03-13
feat(ui): add a filter on disabled state

educlos:
Following https://github.com/distribworks/dkron/issues/912, add a new filter on the jobs page to only show disabled or enabled jobs.

### DIFF
--- a/ui/src/jobs/JobList.tsx
+++ b/ui/src/jobs/JobList.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { 
+import {
     Datagrid,
     TextField,
     NumberField,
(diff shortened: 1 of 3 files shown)


ui/src/jobs/JobList.tsx
```diff
@@ -24,6 +24,10 @@ const JobFilter = (props: any) => (
             { id: 'success', name: 'Success' },
             { id: 'failed', name: 'Failed' },
         ]} />
+        <SelectInput source="disabled" choices={[
```
victorcoder, line 27: Why not use `BooleanInput`? the we could remove this code too https://github.com/distribworks/dkron/pull/923/files#diff-5172cc423c1e538ec337bcdf8abf914469f842f13e0c654a55aed5a58ab93460R175

educlos: True, it is indeed way cleaner

victorcoder (approved):
LGTM I will generate assets files

Issue #913, opened 2021-03-05
Allow to show more than 25 jobs per pages


educlos:
On the jobs page we can select the "Rows per page". The current options are 5, 10 and 25.
It would be nice to add 50 and 100 for instance, for when you have a lot of jobs.

**Additional context**
I checked the code to try to add it myself, but was not successful :disappointed:

yvanoers:
CLosed via #926.

Issue #870, opened 2020-12-18
Dark theme does not survive page reload


alexanderfefelov:
Dkron 3.1.0

Enable the dark theme. Reload (`F5` or `Ctrl+R`) any page. The current theme will be switched to light.

victorcoder:
Yes it seems 🤔 , notice that I'm super noob with React and react admin, I just copied the config from the demo https://marmelab.com/react-admin-demo/#/configuration and it behaves the same, so I need more work to solve this 😅

PR #926, opened 2021-03-18
UI fixes

yvanoers:
Several fixes to the UI, including a fix for #870 and the implementation of #913.
Does not include a new assets_vfsdata.go yet.


ui/src/createAdminStore.js
```diff
@@ -0,0 +1,70 @@
...
+                // our own middlewares here
+            ),
+            // our own enhancers here
+        ),        
+    );
+    sagaMiddleware.run(saga);
+    return store;
+};
+
+export default createAdminStore;
```
victorcoder, line 70:
```suggestion
export default createAdminStore;

```

victorcoder (approved):
Excellent work, thanks!

PR #936, opened 2021-04-05
Improve tests for http executor

sysadmind:
While troubleshooting issues with HTTP jobs, I noticed that test cases that I expected to fail, were not causing the test to fail. I dug in and realized that the errors from the output were not being checked. The returned error was being checked, but this is hard coded to nil in the return so it offers very little value.

This adds some common test cases for the HTTP executor and ensures that many input validations return an error and known good configurations work as expected.

This also updates the Dockerfile to go 1.16 because 1.15 does not include go:embed and was causing `make test` to fail.

New Test Output:
```
--- PASS: TestExecute (0.03s)
    --- PASS: TestExecute/Expected_200 (0.01s)
    --- PASS: TestExecute/Expected_400 (0.00s)
    --- PASS: TestExecute/Expected_404 (0.00s)
    --- PASS: TestExecute/Unexpected_400_is_error (0.00s)
    --- PASS: TestExecute/Empty_URL_is_error (0.00s)
    --- PASS: TestExecute/Empty_method_is_error (0.00s)
    --- PASS: TestExecute/Expected_GET_Response (0.00s)
    --- PASS: TestExecute/Expected_POST_Response (0.02s)
    --- PASS: TestExecute/Unexpected_POST_Response_is_error (0.00s)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.15
+FROM golang:1.16
 LABEL maintainer="Victor Castell <victor@victorcastell.com>"
 
 EXPOSE 8080 8946
--- a/builtin/bins/dkron-executor-http/http_test.go
+++ b/builtin/bins/dkron-executor-http/http_test.go
@@ -1,51 +1,103 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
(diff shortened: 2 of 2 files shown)


victorcoder (approved):
❤️  thx!

Issue #935, opened 2021-03-30
HTTP job is executed by all members


bittrance:
## Describe the bug
All jobs seems to be executed by all nodes. If I understand the log output correctly, the leader is explicitly instructing all nodes to execute the job.

I assume there is some specific circumstance here, but I don't know what that would be?

## To Reproduce
Job definition:
```javascript
{
  name: 'job-0',
  schedule: '@every 2s',
  executor: 'http',
  executor_config: {
    method: 'PUT',
    url: 'http://targeturl/log/job-0',
    timeout: '10',
    expectCode: '200',
    expectBody: 'ok'
  }
}
```

## Expected behavior
I would have expected that the job would be executed once every two seconds.

## Actual behavior
```
[pod/dkron-66f76ffc99-fjfrt/dkron] time="2021-03-30T07:54:41Z" level=info msg="agent: Calling AgentRun" job_name=job-0 node="10.244.0.67:6868"
[pod/dkron-66f76ffc99-fjfrt/dkron] time="2021-03-30T07:54:41Z" level=info msg="agent: Calling AgentRun" job_name=job-0 node="10.244.0.66:6868"
[pod/dkron-66f76ffc99-fjfrt/dkron] time="2021-03-30T07:54:41Z" level=info msg="agent: Calling AgentRun" job_name=job-0 node="10.244.0.65:6868"
[pod/dkron-66f76ffc99-fjfrt/dkron] time="2021-03-30T07:54:41Z" level=info msg="grpc_agent: Starting job" job=job-0 node=dkron-66f76ffc99-fjfrt
[pod/dkron-66f76ffc99-rgfwb/dkron] time="2021-03-30T07:54:41Z" level=info msg="grpc_agent: Starting job" job=job-0 node=dkron-66f76ffc99-rgfwb
[pod/dkron-66f76ffc99-b6tzt/dkron] time="2021-03-30T07:54:41Z" level=info msg="grpc_agent: Starting job" job=job-0 node=dkron-66f76ffc99-b6tzt
[pod/dkron-66f76ffc99-rgfwb/dkron] time="2021-03-30T07:54:43Z" level=info msg="grpc_agent: Starting job" job=job-0 node=dkron-66f76ffc99-rgfwb
[pod/dkron-66f76ffc99-fjfrt/dkron] time="2021-03-30T07:54:43Z" level=info msg="agent: Calling AgentRun" job_name=job-0 node="10.244.0.66:6868"
[...]

PR #938, opened 2021-04-06
Clarify that default is for all nodes run each job

bittrance:
As can be seen in #935, I had not understood that default was for jobs to be executed by all nodes, when there is no tagging to say differently. This PR updates the docs to make this default clear. I believe that if this PR had been applied, I would not have logged the issue in question.


victorcoder (approved):
LGTM thanks!

PR #948, opened 2021-04-28
Fix shell executor crashing on missing command

### LINKED ISSUE #945: Empty Executor Config will cause `invalid memory address or nil pointer dereference` in SHELL Executor
**Describe the bug**
 
 In a new job,  I set `executor=shell`, but forgot to set `executor_config`.
When this job executes, it will cause `invalid memory address or nil pointer dereference` and panics `dkron agent`.

**To Reproduce**
1. add a new job
2. set `executor=shell`
3. leave `executor_config` blank 
4. run this job

**Expected behavior**
Empty Executor Config in a job should do nothing and will not panics `dkron agent`

**Bug**
In `dkron/builtin/bins/dkron-executor-shell/shell.go`, Line 150-176:
```
// Determine the shell invocation based on OS
func buildCmd(command string, useShell bool, env []string, cwd string) (cmd *exec.Cmd, err error) {
	var shell, flag string

	if useShell {
		if runtime.GOOS == windows {
			shell = "cmd"
			flag = "/C"
		} else {
			shell = "/bin/sh"
			flag = "-c"
		}
		cmd = exec.Command(shell, flag, command)
	} else {
		args, err := shellwords.Parse(command)
		if err != nil {
			return nil, err
		}
		cmd = exec.Command(args[0], args[1:]...)
	}
	if env != nil {
		cmd.Env = append(os.Environ(), env...)
	}
	cmd.Dir = cwd
	return
}
```
if `command` is an empty string, then `args` is a empty array. 

`args[0]` and `args[1:]` will out of array bounds and cause `invalid memory address or nil pointer dereference`.

yvanoers:
Fixes #945

### DIFF
--- a/builtin/bins/dkron-executor-shell/shell.go
+++ b/builtin/bins/dkron-executor-shell/shell.go
@@ -165,6 +165,9 @@ func buildCmd(command string, useShell bool, env []string, cwd string) (cmd *exe
 		if err != nil {
 			return nil, err
 		}
+		if len(args) == 0 {
+			return nil, errors.New("shell: Command missing")
+		}
 		cmd = exec.Command(args[0], args[1:]...)
 	}
 	if env != nil {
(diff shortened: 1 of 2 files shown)


victorcoder (approved):
LGTM thx!

PR #950, opened 2021-05-01
Clarify runoncreate and run behaviors

### LINKED ISSUE #946: Additional docs for runoncreate option
**Is your feature request related to a problem? Please describe.**
Jobs created with `runoncreate` seem to both create and execute the job as part of the request and do not return a response until the job execution has completed. Jobs created without `runoncreate` respond immediately after the job has been created.

**Describe the solution you'd like**
Docs should be clear about the execution difference between `runoncreate` and normal jobs (`runoncreate` = blocking job execution (job creation + job execution) VS. **no** `runoncreate` = "asynchronous" job execution (job creation only, execution happens later)). **OR** `runoncreate` should operate asynchronously and return once the job has been created and execute the job in the background, just like adding a scheduled job. 

**Describe alternatives you've considered**
Seems like adding a job and executing immediately without the REST request hanging while the job executes can be done by creating the job as disabled and enabling/running it after creation, however this requires multiple requests, and my initial assumption was that `runoncreate` would replicate this behavior. I have used other schedulers that treat a "run immediately" job exactly the same as a scheduled job which is where my confusion came from.

Appreciate the project! I just had the wrong assumption that `runoncreate` jobs would return the job object and execute the job in the background rather than blocking until execution completed. Took me a while to figure this out so hope it helps someone.

yvanoers:
Fixes #946.

### DIFF
--- a/website/content/v2.0/swagger.yaml
+++ b/website/content/v2.0/swagger.yaml
@@ -67,7 +67,10 @@ paths:
             $ref: "#/definitions/job"
         - in: query
           name: runoncreate
-          description: If present, regardless of any value, causes the job to be run immediately after being succesfully created or updated.
+          description: |
+            If present, regardless of any value, causes the job to be run immediately after being succesfully created or updated.
+
+            This will cause the server to wait until the job has finished before returning a result.
           required: false
           schema:
             type: boolean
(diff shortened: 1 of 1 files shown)


victorcoder (approved):
LGTM thx!

Issue #951, opened 2021-05-06
Many race conditions


sysadmind:
**Describe the bug**
Recently I have been looking through the code of dkron to help diagnose some of our issues:
- No leader elected indefinitely
- Leader kicks off job and agent can not look up job details

During my investigations I have tried running the test suite. I have race detection turned on by default and this brought up several race conditions.

**To Reproduce**
1. From the root directory of the project, execute `go test -timeout 30s github.com/distribworks/dkron/v3/dkron -count=1 -race`

**Expected behavior**
Tests pass with no race conditions found

** Specifications:**
 - OS: Any
 - Version: git hash 0de02f70c247c32816d6c7b76fe9461ab3b8d10e (v3.1.6)

**Additional context**
Some of these may be a test suite parallelization symptom, but ultimately I believe all of the functions should be more concurrency safe. One such example is `dkron/log.go` where log is a package scoped var so any time `InitLogger` is called, that log value gets overwritten. An alternative approach would be for functions that need a logger to take the logger as a parameter ( a more pure function/functional programming approach). That would prevent needing a package scoped var that is so susceptible to race conditions in general. 

I would be happy to help on this effort if there is support of changing function signatures to accept the logger or other resources that may be involved in a data race.

victorcoder:
Refactor logging is something that I wanted to do since long ago, a PR will be very welcome, thanks.

PR #963, opened 2021-05-10
Overhaul logging to avoid package level var

sysadmind:
This converts all logging calls to a logrus instance that is passed into types or functions instead of a package level variable. This is to help prevent any possible race conditions with the variable and it also helps show the dependency on logging for each type instead of an implicit reference.

Originally I attempted to abstract a logging interface away from logrus but so much of the logrus API is being used that it was quite difficult to abstract away and there was not much direct benefit from doing so. The tests are all passing and I did test running docker-compose locally. I can add jobs via the UI and execute them with no panics. I feel relatively confident with these changes, but I'm not sure how complete the test coverage is in order to prove that.

I'm open to feedback here if there are improvements that can be made. The Job logging calls felt the most difficult because the Run method is not allowed to have parameters so I had to have the scheduler set the logger on the jobs before returning them. 

This helps improve #951

### DIFF
--- a/cmd/leave.go
+++ b/cmd/leave.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/distribworks/dkron/v3/dkron"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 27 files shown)


victorcoder (approved):
Awesome thx!

PR #967, opened 2021-05-17
Fix race conditions

### LINKED ISSUE #951: Many race conditions

sysadmind:
- Add workaround for package level variables in gin framework
- Add mutext around Scheduler.started to prevent concurrent modifications
- Scheduler.Started renamed to .started as a private property, Ability to query on started status now through Started()

These changes allow the tests to be run with the -race flag and not return any warnings about race conditions.

Fixes #951

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -244,7 +244,7 @@ func (a *Agent) Stop() error {
 		a.Store.Shutdown()
 	}
 
-	if a.config.Server && a.sched.Started {
+	if a.config.Server && a.sched.Started() {
 		a.sched.Stop()
 		a.sched.ClearCron()
 	}
--- a/dkron/log.go
+++ b/dkron/log.go
@@ -2,10 +2,15 @@ package dkron
 
 import (
 	"io/ioutil"
+	"sync"
 
 	"github.com/gin-gonic/gin"
 	"github.com/sirupsen/logrus"
 )
+
+// ginOnce is a wrapper around gin global var changes. This is a workaround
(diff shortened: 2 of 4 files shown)


sysadmind:
This test seems to intermittently fail now for me which makes me believe that it is a result of these changes, but it is not very clear as to why. I'll dig in further as I have time.

sysadmind:
close/reopen to trigger CI

PR #968, opened 2021-05-20
Refactor processFilteredNodes for testing

sysadmind:
I was digging into the intermittent failures of Test_processFilteredNodes and had trouble really understanding what the function, and the complimenting filterNodes were doing so I broke some of the logic out into smaller functions to encapsulate the logic and make testing of each portion easier. I added unit tests for these smaller functions to help catch potential issues with each. I left Test_processFilteredNodes as it serves well as a higher level test of all the other funtions.

This adjusts the thresholds for the even distribution of picking nodes when the cardinality is lower than the possible nodes. Due to the randomness, it is quite hard to test how evenly distributed the values are within the original thresholds. This would cause many intermittent test failures when the values were just over or under the thresholds.

This also moves the seed of math.Rand into the agent startup. Reseeding the randomness every function call seems like extra overhead for no value. I do not believe that constantly reseeding the randomness increases randomness in anyway.

This may help the tests for #967

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -745,14 +745,14 @@ func (a *Agent) processFilteredNodes(job *Job) (map[string]string, map[string]st
 	return nodes, tags, nil
 }
 
-// filterNodes determines which of the execNodes have the given tags
+// filterNodes determines which of the provided nodes have the given tags
 // Returns:
-// * the map of execNodes that match the provided tags
+// * the map of allNodes that match the provided tags
 // * a clean map of tag values without cardinality
 // * cardinality, i.e. the max number of nodes that should be targeted, regardless of the
 //   number of nodes in the resulting map.
 // * an error if a cardinality was malformed
-func filterNodes(execNodes map[string]serf.Member, tags map[string]string) (map[string]serf.Member, map[string]string, int, error) {
+func filterNodes(allNodes map[string]serf.Member, tags map[string]string) (map[string]serf.Member, map[string]string, int, error) {
 	ct, cardinality, err := cleanTags(tags)
 	if err != nil {
 		return nil, nil, 0, err
@@ -761,7 +761,7 @@ func filterNodes(execNodes map[string]serf.Member, tags map[string]string) (map[
 	matchingNodes := make(map[string]serf.Member)
 
 	// Filter nodes that lack tags
-	for name, member := range execNodes {
+	for name, member := range allNodes {
 		if nodeMatchesTags(member, ct) {
 			matchingNodes[name] = member
 		}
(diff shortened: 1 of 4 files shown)


dkron/agent.go
```diff
@@ -745,51 +746,34 @@ func (a *Agent) processFilteredNodes(job *Job) (map[string]string, map[string]st
...
-			tagCardinality, err := strconv.Atoi(tc[1])
-			if err != nil {
-				return nil, nil, 0, err
-			}
-			if tagCardinality < cardinality {
-				cardinality = tagCardinality
-			}
+	for name, member := range execNodes {
+		nodeMatches := nodeMatchesTags(member, ct)
+		if nodeMatches {
```
yvanoers, line 766: wouldn't
`if nodeMatchesTags(...) {`
be equally readable?

yvanoers (commented):
Nice work, certainly more readable.
Though I do suspect this is not as fast, but it would have to be measured to be sure and assess whether or not difference is significant, especially for larger clusters. This concern is the reason why I didn't pull it apart when last I touched this.

yvanoers:
I didn't realize the intermittent failures of the filterNodes test were frequent enough to make someone investigate it. I will intensify my efforts to fix that.

sysadmind:
Thanks for the review @yvanoers. I incorporated your feedback in another commit. 

I'm not sure how many people are regularly running the tests to catch the intermittent failures, but I have been running them a ton recently because of my work on race conditions in tests. I noticed in another one of your PRs that you closed and reopened it after the same failing test that I had in before. That leads me to believe that there are still some flaky tests beyond what this PR intends to solve.

yvanoers:
> That leads me to believe that there are still some flaky tests beyond what this PR intends to solve.

Oh yes, there definitely is. I know exactly where the problem lies and I have a solution waiting in a branch (as sort-of promised a couple days ago). But it would conflict with this PR. If this PR gets merged I can go ahead and finish my changes. Or if it gets rejected, there are a couple of changes in this PR I'd like to incorporate.

dkron/agent.go
```diff
@@ -745,51 +746,33 @@ func (a *Agent) processFilteredNodes(job *Job) (map[string]string, map[string]st
...
-// Out param! The incoming execNodes map is modified.
 // Returns:
-// * the (modified) map of execNodes
-// * a map of tag values without cardinality
+// * the map of execNodes that match the provided tags
+// * a clean map of tag values without cardinality
 // * cardinality, i.e. the max number of nodes that should be targeted, regardless of the
 //   number of nodes in the resulting map.
 // * an error if a cardinality was malformed
 func filterNodes(execNodes map[string]serf.Member, tags map[string]string) (map[string]serf.Member, map[string]string, int, error) {
```
victorcoder, line 755: Maybe `execNodes` -> `allNodes`? but I'm nitpicking

victorcoder (dismissed):
Good work here, thanks a lot!

sysadmind:
I think that's a more readable name @victorcoder. I just pushed the update. Thanks!

victorcoder (approved):

PR #970, opened 2021-05-23
Add @minutely custom schedule again

### LINKED ISSUE #969: Predefined schedule "@minutely" does not work
**Describe the bug**
Despite mentioned in the documentation here (https://dkron.io/v2.0/usage/cron-spec/) the @minutely predefined schedule is not supported.

**To Reproduce**
1. Try to create a job with @minutely schedule.

**Actual behavior**
Job is not created, error returned.

**Expected behavior**
Job created, runs every 60 seconds.

** Specifications:**
 - Version 3.1.6

**Additional context**
As I can see, https://github.com/robfig/cron/blob/master/parser.go#L365 (library used to parse the cron specs) clearly does not have the @minutely support (I relate to the main version, but it's the same for the one used in Dkron 3.1.6)

So it looks like a bug in the documentation.

yvanoers:
Fixes #969

### DIFF
--- a/extcron/extparser.go
+++ b/extcron/extparser.go
@@ -21,10 +21,14 @@ func NewParser() cron.ScheduleParser {
 
 // Parse parses a cron schedule specification. It accepts the cron spec with
 // mandatory seconds parameter, descriptors and the custom descriptors
-// "@at <date>" and "@manually".
+// "@at <date>", "@manually" and "@minutely".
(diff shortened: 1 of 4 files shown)


yvanoers:
Closing & reopening to trigger tests

victorcoder (approved):
LGTM Thanks!

PR #973, opened 2021-05-28
Add build target + improve test output

yvanoers:
Adds a make target for 'main' that rebuilds the main binary (and plugins) and any intermediates when necessary.
Colorize PASS or FAIL in test output to be able to easily spot the failures.


victorcoder (approved):
LGTM

PR #975, opened 2021-06-06
Convert shell plugin to internal plugin

vcastellm:
The purpose of this PR is to embed the shell plugin in the main dkron binary, that will facilitate creating a single binary with the most important executor, for easy deployment using a single binary.


cmd/plugins.go
```diff
@@ -107,10 +112,11 @@ func getPluginName(file string) (string, bool) {
 	return name, true
 }
 
-func (p *Plugins) pluginFactory(path string, pluginType string) (interface{}, error) {
+func (p *Plugins) pluginFactory(path string, args []string, pluginType string) (interface{}, error) {
 	// Build the plugin client configuration and init the plugin
 	var config plugin.ClientConfig
-	config.Cmd = exec.Command(path)
+	fmt.Println(path)
```
yvanoers, line 118: Debugging remnant?

yvanoers (changes requested):

yvanoers (approved):
LGTM!

PR #976, opened 2021-06-06
Replace boltdb with fast-store

vcastellm:
Replacing BoltDB with fast-store increased the speed of the Raft log operations enormously, allowing the user to choose the degree of duration in disk.

Under my tests, with the same degree of duration as boltdb (High) the speed is x2 but using the Mid duration, the default that sync to disk every second, the speed increases tenfold.

Considering that Dkron is not the source of truth for the data it stores, the cost-benefit of using less durability is high.

Note: This will be included in the upcoming v4.x


dkron/agent.go
```diff
@@ -312,7 +312,7 @@ func (a *Agent) setupRaft() error {
 		}
 
 		// Create the BoltDB backend
```
yvanoers, line 314: You missed updating the comment here.

yvanoers (changes requested):
Besides the comment, LGTM.
But does this change require users to do some kind of migration, or will this work transparently?

victorcoder:
Yes @yvanoers this is a breaking change, this is why this will be published in 4.x, you can not upgrade in place but it only requires of a rolling deploy.

yvanoers (approved):
LGTM!

yvanoers:
> Yes @yvanoers this is a breaking change, this is why this will be published in 4.x, you can not upgrade in place but it only requires of a rolling deploy.

Ok; is there, or will there be documentation on how to do that? I'm unaware of any.

victorcoder:
@yvanoers doc for upgrade strategies here https://dkron.io/usage/upgrade/#rolling-upgrade

yvanoers:
Ah yes, great!
And here's a PR to improve some of the text: #980 :)

vcastellm:
Superseded by #1172

PR #982, opened 2021-06-11
Fix intermittent test false negative

yvanoers:
This fixes the test of processFilteredNodes that sometimes failed.
This became a large refactor:
* processFilteredNodes no longer exists and is replaced by several smaller functions
* significantly revised test code
* allows for injecting the selector, which paves the way for e.g. selecting nodes based on weight
* some code improvements and markup revisions I came across

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -98,14 +98,8 @@ WAIT:
 		goto WAIT
 	}
 
-	// Check if we should do a graceful leave
-	graceful := false
-	if sig == syscall.SIGTERM || sig == os.Interrupt {
-		graceful = true
-	}
-
 	// Fail fast if not doing a graceful leave
-	if !graceful {
+	if sig != syscall.SIGTERM && sig != os.Interrupt {
 		return 1
 	}
 
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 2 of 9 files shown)


cmd/agent.go
```diff
@@ -98,14 +98,8 @@ WAIT:
...
 
-	// Check if we should do a graceful leave
-	graceful := false
-	if sig == syscall.SIGTERM || sig == os.Interrupt {
-		graceful = true
-	}
-
 	// Fail fast if not doing a graceful leave
-	if !graceful {
+	if sig != syscall.SIGTERM && sig != os.Interrupt {
```
victorcoder, line 102: All signals except SIGTERM or interrupt are becoming graceful, why?

dkron/run.go
```diff
@@ -28,35 +28,37 @@ func (a *Agent) Run(jobName string, ex *Execution) (*Job, error) {
 
 	// In the first execution attempt we build and filter the target nodes
 	// but we use the existing node target in case of retry.
-	var filterMap map[string]string
+	var targetNodes []serf.Member
```
victorcoder, line 31: Refactor to `Node` type?

victorcoder (changes requested):

cmd/agent.go
yvanoers: No, only SIGTERM and interrupt are graceful. For any other signal, the condition evaluates to true, because it is neither SIGTERM nor interrupt and then `return 1` causes non-graceful termination.
The code is equivalent.

dkron/run.go
yvanoers: Done

dkron/agent.go
```diff
@@ -594,33 +594,6 @@ func (a *Agent) IsLeader() bool {
...
-	return a.serf.Members()
-}
-
-// LocalMember is used to return the local node
-func (a *Agent) LocalMember() serf.Member {
-	return a.serf.LocalMember()
-}
-
-// Leader is used to return the Raft leader
-func (a *Agent) Leader() raft.ServerAddress {
```
victorcoder, line 605: Remember that dkron is also used as library from the Pro version. This is used.

dkron/agent.go
```diff
@@ -594,33 +594,6 @@ func (a *Agent) IsLeader() bool {
...
-	return a.raft.Leader()
-}
-
-// Servers returns a list of known server
-func (a *Agent) Servers() (members []*ServerParts) {
```
victorcoder, line 610: Used too

victorcoder (changes requested):
I wouldn't remove any other public method, they could be in use by third parties.

yvanoers:
Whoops, didn't realize, sorry. I reverted the removal of seemingly unused code.

victorcoder (approved):
LGTM

PR #984, opened 2021-06-14
Remove old UI

vcastellm:
It will be removed in 4.x


yvanoers (approved):
Yep, was wondering when this would be removed :)
LGTM

PR #988, opened 2021-06-16
Release a light image tag

vcastellm:
Omit all plugins except the shell plugin that will be included in the main binary.


yvanoers (approved):
Looks ok to me

PR #997, opened 2021-06-24
swagger assets from cdn

FedeBev:
Get (updated) Swagger assets from CDN, this way we don't have to keep them updated anymore and the repo is a bit lighter.

Hope this is welcome

Thanks


victorcoder:
Thanks for the contribution @FedeBev but the enterprise users won't be happy with this, some of them use proxied networking environments, and doesn't allow internet connections from internal applications, I prefer to have a bit heavier package than force them to allow this connection from their proxy config.

victorcoder (approved):
LGTM

FedeBev:
> Thanks for the contribution @FedeBev but the enterprise users won't be happy with this, some of them use proxied networking environments, and doesn't allow internet connections from internal applications, I prefer to have a bit heavier package than force them to allow this connection from their proxy config.

Are u sure about that? This just affects the public website (or so i thought)

victorcoder (approved):

PR #1000, opened 2021-07-02
fix comment typo

KevinWu0904:


dkron/config.go
```diff
@@ -81,7 +81,7 @@ type Config struct {
 	AdvertiseRPCPort int `mapstructure:"advertise-rpc-port"`
 
 	// LogLevel is the log verbosity level used.
-	// It cound be (debug|info|warn|error|fatal|panic).
+	// It could be (debug|info|warn|error|fatal|panic).
```
yvanoers, line 84: It is supposed to be 'can', actually. 'Could' is hypothetical, which doesn't apply in this case.

yvanoers (changes requested):
Thanks, but let's fix it properly :)

KevinWu0904:
@yvanoers Thanks for your review :)

yvanoers (approved):

yvanoers:
Thanks @KevinWu0904 !

PR #1006, opened 2021-07-06
use relative API_URL

fopina:
Similar to a PR I've done in a past version (v1 already I think) for the old UI, this would allow to serve dkron behind a reverse proxy on a custom path.

I've gone through the code (not extensively) looking for places where DKRON_API_URL would be loaded from a different path but I couldn't find any (mostly as it's all react now).

### DIFF
--- a/dkron/ui.go
+++ b/dkron/ui.go
@@ -81,7 +81,7 @@ func (h *HTTPTransport) UI(r *gin.RouterGroup) {
 				ln = l.Name
 			}
 			ctx.HTML(http.StatusOK, "index.html", gin.H{
-				"DKRON_API_URL":          fmt.Sprintf("/%s", apiPathPrefix),
+				"DKRON_API_URL":          fmt.Sprintf("../%s", apiPathPrefix),
 				"DKRON_LEADER":           ln,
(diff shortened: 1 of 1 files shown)


dkron/ui.go
```diff
@@ -81,7 +81,7 @@ func (h *HTTPTransport) UI(r *gin.RouterGroup) {
 				ln = l.Name
 			}
 			ctx.HTML(http.StatusOK, "index.html", gin.H{
-				"DKRON_API_URL":          fmt.Sprintf("/%s", apiPathPrefix),
+				"DKRON_API_URL":          fmt.Sprintf("../%s", apiPathPrefix),
```
victorcoder, line 84: Is this working serving the API from a root path?

fopina: It should as UI is served from /ui but I’ll double check

victorcoder: Can you confirm?

fopina: sorry, I forgot but yes, accessing directly (on root path) still works as expected

victorcoder (approved):

PR #1009, opened 2021-07-12
bugfix: fd-leak when RPC client communication fail

thierry-f-78:
in function 'ExecutionDone' file 'dkron/grpc_client.go', when the process
fails, it use a code path without 'conn.Close()' directive. It seems an
error because all code path closes the connexion even when succes is
reported.

### DIFF
--- a/dkron/grpc_client.go
+++ b/dkron/grpc_client.go
@@ -78,13 +78,13 @@ func (grpcc *GRPCClient) ExecutionDone(addr string, execution *Execution) error 
 		}).Error("grpc: error dialing.")
 		return err
 	}
+	defer conn.Close()
 
 	d := proto.NewDkronClient(conn)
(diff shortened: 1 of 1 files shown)


jjsaunier:
Nice catch

If the connection should be terminated when function returned, you can tell to go to do it in one line and for every case, as soon as  the conn is available `defer conn.Close()` and remove all `conn.Close()` from every possible cases, will be more reliable and less complex

```
defer conn.Close()
d := proto.NewDkronClient(conn)
```

thierry-f-78:
You're right. it's done.

yvanoers (commented):
Nice work! LGTM!

yvanoers (approved):

victorcoder (approved):
Nice one, thanks!

PR #1010, opened 2021-07-18
fix: wait all jobs done

KevinWu0904:
Signed-off-by: KevinWu0904 <KevinWu0904@outlook.com>


KevinWu0904:
Actually, the underlaying robig/cron lib's stop function will return a context for developers to wait. So if we want to stop all jobs gracefully, maybe we should wait the context done. Correct me if any problems, thanks!

image

yvanoers:
What problem does this solve?

Issue #1011, opened 2021-08-12
Update the status text in dashboard page 


PHChenGit:
Hi dkron,

I found the legacy code about nodes status text mapping in `static/js/dashboard.js`, and the code translate the status code to  human readable text.  the legacy code looks like below.

<img width="668" alt="Screen Shot 2021-08-12 at 3 17 24 PM" src="https://user-images.githubusercontent.com/16322329/129155059-c638dae7-2554-43eb-a68b-61856d3e2beb.png">

I think it is great.

However, the status text is not translated to human readable text in the latest version. I guest it just missing the mapping in the react-admin. So, I make some change to the code

PHChenGit:
Here is the PR.

https://github.com/distribworks/dkron/pull/1012

PR #1012, opened 2021-08-12
feat: Update node status text

PHChenGit:
Issue: #1011

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -365,13 +365,14 @@ type MId struct {
 	serf.Member
 
 	Id string `json:"id"`
+	StatusText string `json:"statusText"`
 }
 
 func (h *HTTPTransport) membersHandler(c *gin.Context) {
 	mems := []*MId{}
 	for _, m := range h.agent.serf.Members() {
 		id, _ := uuid.GenerateUUID()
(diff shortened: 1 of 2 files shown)


yvanoers:
@PHChenGit 
Thanks for your efforts!
The change looks sound by itself, but I'd prefer it if the frond end would be fed the names of the statuses by the backend. Reason for this is that the names of the statuses originate from the external Serf library and this way there's no duplication of externally defined names.
Should be doable in 3 or 4 lines of code in api.go and Dashboard.tsx.

Would you be willing to rework it like this?

PHChenGit:
Hi @yvanoers 

Yes, I can do it.
I pushed some new code to this PR and I found the Serf library has status to string method.
I also hope to avoid to change the `serf.member` struct, so I add `statusText` to struct `MId`

After changed the code, the ui is looked like below. You can focus on the `Status` column
First, It's a screenshot before changed.
Screen Shot 2021-08-13 at 3 45 51 PM


And, It's a screenshot after changed.
Screen Shot 2021-08-13 at 4 19 35 PM

yvanoers (approved):
LGTM!

PR #1017, opened 2021-09-02
fix: tags role was not work

wayne666:


wayne666 (commented):
When i use role tag in dkron.yml, it's not worked. So i fix it. Plz review the commit. Thanks

victorcoder:
Sorry for this not being clear in the documentation but the `role` tag is reserved for internal usage https://dkron.io/upgrading/from_v1_to_v2/#changing-tags-and-metadata

PR #1025, opened 2021-11-09
truncate execution output in default view

fopina:
`maxBufSize = 256000` in the shell handler, 100 executions per job, 25M to download when listing executions (and added browser rendering time)

The idea of this PR is to add an API parameter to truncate execution output but leave default to not truncate and avoid API breaking changes.

Then update UI to load list truncated but allow individual executions to have output fully loaded.

## test

Create job with shell command `seq 1000000` (`seq 1000000 | wc -c` = `6888894`)

```
curl 'http://localhost:8080/v1/jobs' \
     --data-raw '{"name":"test","schedule":"@manually","executor":"shell","executor_config":{"command":"seq 1000000"}}'
```

Run it 100+ times:

```
for i in $(seq 100); do curl 'http://localhost:8080/v1/jobs/test' -X POST; done
```

## Notes

Some parts pending re-design as I'm a complete react ignorant:

* This PR does not include the generated assets to allow for a cleaner review first
* `output_size` was added to dataprovider though it doesn't look correct there - how to send extra query param (`output_size`) from the ReferenceManyField component without modifying dataprovider? Or modifying it in a more generic way that makes sense to other components/resources
* `SpecialOutputPanel` hammered in to have the `load full output` button, any suggestion for better UI/code? also need(should?) hide button if output is not truncated
* missing tests for the new api endpoint (left for after initial review/comments)

### DIFF
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -119,7 +119,7 @@ func (h *HTTPTransport) APIRoutes(r *gin.RouterGroup, middleware ...gin.HandlerF
 	// Place fallback routes last
 	jobs.GET("/:job", h.jobGetHandler)
 	jobs.GET("/:job/executions", h.executionsHandler)
-	jobs.GET("/:job/execution/:execution", h.executionHandler)
+	jobs.GET("/:job/executions/:execution", h.executionHandler)
 }
 
 // MetaMiddleware adds middleware to the gin Context.
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -329,6 +329,11 @@ func (h *HTTPTransport) restoreHandler(c *gin.Context) {
 	renderJSON(c, http.StatusOK, string(resp))
 }
 
+type apiExecution struct {
+	*Execution
+	OutputTruncated bool `json:"output_truncated"`
+}
+
 func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 	jobName := c.Param("job")
 
@@ -337,9 +342,9 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 		sort = "started_at"
 	}
 	order := c.DefaultQuery("_order", "DESC")
-	output_size_limit, err := strconv.Atoi(c.DefaultQuery("output_size_limit", ""))
+	outputSizeLimit, err := strconv.Atoi(c.DefaultQuery("output_size_limit", ""))
 	if err != nil {
-		output_size_limit = -1
+		outputSizeLimit = -1
 	}
 
 	job, err := h.agent.Store.GetJob(jobName, nil)
@@ -362,18 +367,25 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 		return
 	}
 
-	if output_size_limit > 0 {
-		// truncate execution output
-		for _, execution := range executions {
-			size := len(execution.Output)
-			if size > output_size_limit {
-				execution.Output = execution.Output[size-output_size_limit:]
+	apiExecutions := make([]*apiExecution, len(executions))
+	for j, execution := range executions {
+		apiExecutions[j] = &apiExecution{execution, false}
+		if outputSizeLimit > -1 {
+			apiExecutions[j].OutputTruncated = true
+			// truncate execution output
+			if outputSizeLimit == 0 {
+				apiExecutions[j].Output = ""
+			} else {
+				size := len(execution.Output)
+				if size > outputSizeLimit {
+					apiExecutions[j].Output = apiExecutions[j].Output[size-outputSizeLimit:]
+				}
 			}
 		}
 	}
 
 	c.Header("X-Total-Count", strconv.Itoa(len(executions)))
-	renderJSON(c, http.StatusOK, executions)
+	renderJSON(c, http.StatusOK, apiExecutions)
 }
 
 func (h *HTTPTransport) executionHandler(c *gin.Context) {
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -337,9 +337,9 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 		sort = "started_at"
 	}
 	order := c.DefaultQuery("_order", "DESC")
-	output_size, err := strconv.Atoi(c.DefaultQuery("output_size", ""))
+	output_size_limit, err := strconv.Atoi(c.DefaultQuery("output_size_limit", ""))
 	if err != nil {
-		output_size = -1
+		output_size_limit = -1
 	}
 
 	job, err := h.agent.Store.GetJob(jobName, nil)
@@ -362,12 +362,12 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 		return
 	}
 
-	if output_size > 0 {
+	if output_size_limit > 0 {
 		// truncate execution output
 		for _, execution := range executions {
 			_s := len(execution.Output)
-			if _s > output_size {
-				execution.Output = execution.Output[_s-output_size:]
+			if _s > output_size_limit {
+				execution.Output = execution.Output[_s-output_size_limit:]
 			}
 		}
 	}
--- a/dkron/api.go
+++ b/dkron/api.go
@@ -365,9 +365,9 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 	if output_size_limit > 0 {
 		// truncate execution output
 		for _, execution := range executions {
-			_s := len(execution.Output)
-			if _s > output_size_limit {
-				execution.Output = execution.Output[_s-output_size_limit:]
+			size := len(execution.Output)
+			if size > output_size_limit {
+				execution.Output = execution.Output[size-output_size_limit:]
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


vcastellm (dismissed):
Sorry for the delay @fopina LGTM thanks!

dkron/api.go
```diff
@@ -357,14 +362,55 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 		return
 	}
 
+	if output_size > 0 {
```
yvanoers, line 365: The fallback for output_size is -1 according to code above.
This check also falls back to returning everything when output_size is 0. Is that intended?

dkron/api.go
```diff
@@ -357,14 +362,55 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
...
+		// truncate execution output
+		for _, execution := range executions {
+			_s := len(execution.Output)
```
yvanoers, line 368: Why the leading `_` in `_s` ? Why not simply use `size` as the variable name ?

ui/src/jobs/JobShow.tsx
```diff
@@ -32,6 +36,48 @@ const SuccessField = (props: any) => {
...
+            disabled={loading}
+        >
+            <FullIcon/>
+        </Button>
+    );
+};
+
+const SpecialOutputPanel = ({ id, record, resource }: any) => {
+    // FIXME: hide button if not required
+    return (<div className="execution-output"><div><FullButton record={record} /></div>{record.output || "Empty output"}</div>);
```
yvanoers, line 78: I'd like to see that if the output was truncated, this is shown to the user, so they know which part is not shown (the tail is shown as opposed to the head).

yvanoers (changes requested):

yvanoers:
Forgot to mention in the review: could you add tests to the new features in api.go?

ui/src/jobs/JobShow.tsx
fopina: I would like that too as well, as hiding the show button, but as the FIXME says, I couldn’t get around to do it, open to help :)

yvanoers: I think it is necessary to have the API return the size of the output when it is truncated, so the UI can determine whether or not everything is displayed. It doesn't have enough information as it stands.
Can't promise I'll have time to help, but I'll see what I can do :)

fopina:
> Forgot to mention in the review: could you add tests to the new features in api.go?

this is one of the original notes I left in the PR text :) I didn't write the tests as I wasn't sure if the API would remain as it, due to my other notes which are highlighting parts I don't like in the PR.
One of them was hiding the button to highlight when there is anything more (that you also mentioned) but I don't know how the syntax to have those conditions in react, need to revisit it (though I was waiting for specific hints :( )

But also had a note for the way `output_size` is handled currently as I'm sure there's a prettier way...

yvanoers:
> this is one of the original notes I left in the PR text :) I didn't write the tests as I wasn't sure if the API would remain as it, due to my other notes which are highlighting parts I don't like in the PR.

Ah, indeed. I missed that, sorry!
I think the API is fine like this. Regardless of how the front end needs to be built, the ability to limit the output in the executions call is a good improvement. And retrieving data for one execution is also fine - except that I think that endpoint should be called `.../executions/:execution` (plural), come to think of it.

> One of them was hiding the button to highlight when there is anything more (that you also mentioned) but I don't know how the syntax to have those conditions in react, need to revisit it (though I was waiting for specific hints :( )

I can't provide any hints on doing this off the top of my head, I would need to look into that just like you.

dkron/api.go
```diff
@@ -119,6 +119,7 @@ func (h *HTTPTransport) APIRoutes(r *gin.RouterGroup, middleware ...gin.HandlerF
 	// Place fallback routes last
 	jobs.GET("/:job", h.jobGetHandler)
 	jobs.GET("/:job/executions", h.executionsHandler)
+	jobs.GET("/:job/execution/:execution", h.executionHandler)
```
vcastellm, line 122: As @yvanoers said, this will maintain the convention

```suggestion
	jobs.GET("/:job/executions/:execution", h.executionHandler)
```

dkron/api.go
```diff
@@ -336,6 +337,10 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
 		sort = "started_at"
 	}
 	order := c.DefaultQuery("_order", "DESC")
+	output_size, err := strconv.Atoi(c.DefaultQuery("output_size", ""))
```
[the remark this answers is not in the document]
fopina: renamed 👍 (leaving unresolved for verification, not sure what's the project etiquette)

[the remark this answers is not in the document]
fopina: I did it the same way as the other query params (such as `_sort` right before). behavior is if invalid, set to default, simply did the same for consistency.
no worries changing it but I do feel consistency is key

dkron/api.go
fopina: yes, it was intended as I thought it make no sense to allow 0 output_size.
But thinking of it now, maybe it does make sense to support "no output" / `limit=0` ? what do you think?

dkron/api.go
fopina: ✅

ui/src/jobs/JobShow.tsx
fopina: besides the react part to hide or display the button when it should, I'm not even sure where to add said flag in the response.

if I add it to execution struct it will be serialized in the store (with default value) without any real purpose.
any ideas?
I'll give it a try with a `APIExecution` struct (that embeds current Execution)

fopina: haven't modified the UI yet but updated the matching API endpoint using a new struct `apiEndpoint` which includes `output_truncated` flag.
don't really like the fact that every execution data is copied though..
one other option I can think off is to pass the output size limit itself (not whether it's truncated) and just check output size versus limpt in the UI, doesn't sound ideal though..

dkron/api.go
fopina: changed it in the latest commit :)

ui/src/jobs/JobShow.tsx
fopina: managed to make use of that variable neatly in the react code, IMO :)

fopina:
addressed all reviews and also included the test 🎉

fopinappb:
@yvanoers any chance this can be merged?

yvanoers:
Have seen your comment - will get back to you.

fopina:
@yvanoers :)

dkron/api.go
```diff
@@ -357,14 +367,63 @@ func (h *HTTPTransport) executionsHandler(c *gin.Context) {
...
 	}
 
+	apiExecutions := make([]*apiExecution, len(executions))
+	for j, execution := range executions {
+		apiExecutions[j] = &apiExecution{execution, false}
+		if outputSizeLimit > -1 {
+			// truncate execution output
+			if outputSizeLimit == 0 {
+				apiExecutions[j].Output = ""
+				apiExecutions[j].OutputTruncated = true
```
yvanoers, line 377: When len(execution.Output) == 0 && outputSizeLimit == 0, OutputTruncated should be false.
Is this special case for outputSizeLimit==0 particularly useful? Seems like it would work fine with just the bit currently in the else block.

ui/src/jobs/JobShow.tsx
```diff
@@ -32,6 +36,53 @@ const SuccessField = (props: any) => {
...
+            <FullIcon/>
+        </Button>
+    );
+};
+
+const SpecialOutputPanel = ({ id, record, resource }: any) => {
+    return (
+        <div className="execution-output">
+            {record.output_truncated ? <div><FullButton record={record} /></div> : ""}
+            {record.output || "Empty output"}
```
yvanoers, line 81: This will display "Empty output" as well when outputsizelimit=0 but the actual output size is > 0.
A minor thing, but still.

yvanoers (changes requested):

yvanoers:
> @yvanoers :)
Yes, sorry, have a lot on my plate atm.
Anyway, nice work! Couple of findings still, but those should not be a big issue.

dkron/api.go
fopina: you're right, not exactly sure why I added that condition. might have been related to the initial "hacky" implementation..
or maybe it was trying to save a `len()` operation whenever output is not wanted at all

ui/src/jobs/JobShow.tsx
fopina: hmm not sure what to show then. if it's truncated to 0 or it has no output, there's nothing to show there (and I assume the origin `empty output` message is to avoid leaving a empty box which looks "unfinished").

truncated or actually 0 would already be reflected by the `FullButton` presence. Maybe just changing the text to `Nothing to show` makes more sense, rather than adding some condition and alternate text?

fopina:
> Yes, sorry, have a lot on my plate atm. Anyway, nice work! Couple of findings still, but those should not be a big issue.

no worries, I ping just to remind not to complain.

made commits to address the two comments, let me know if you agree

yvanoers (approved):
Stamp of approval!

vcastellm (approved):
LGTM thanks!

PR #1037, opened 2022-01-13
Allow multiple brokers in kafka executor

Espina2:
Kafka is supposed to be distributed, passing only broker can make messages fail if for some reason the broker is not up. In order to circumvent that issue, we can use multiple brokers. This can also improve the publishing rate.

Signed-off-by: Paulo Moura <itsme@paulomoura.com.pt>

### DIFF
--- a/builtin/bins/dkron-executor-kafka/kafka.go
+++ b/builtin/bins/dkron-executor-kafka/kafka.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"log"
+	"strings"
 
 	"github.com/Shopify/sarama"
 	"github.com/armon/circbuf"
@@ -65,7 +66,7 @@ func (s *Kafka) ExecuteImpl(args *dktypes.ExecuteRequest) ([]byte, error) {
 	config.Producer.Return.Successes = true
 	config.Producer.Return.Errors = true
 
-	brokers := []string{args.Config["brokerAddress"]}
+	brokers := strings.Split(args.Config["brokerAddress"], ",")
 	producer, err := sarama.NewSyncProducer(brokers, config)
 	if err != nil {
 		// Should not reach here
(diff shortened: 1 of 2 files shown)


Espina2:
@vcastellm can you take look?

builtin/bins/dkron-executor-kafka/kafka.go
```diff
@@ -65,7 +66,7 @@ func (s *Kafka) ExecuteImpl(args *dktypes.ExecuteRequest) ([]byte, error) {
 	config.Producer.Return.Successes = true
 	config.Producer.Return.Errors = true
 
-	brokers := []string{args.Config["brokerAddress"]}
+	brokers := strings.Split(args.Config["brokerAddress"], ",")
```
yvanoers, line 69: What if people accidentally use an additional space after the comma? Does `NewSyncProducer` handle that properly or should it be taken into account here?

website/content/usage/executors/kafka.md
```diff
@@ -22,7 +22,7 @@ Example
 ```json
 "executor": "kafka",
 "executor_config": {
-    "brokerAddress": "localhost:9092",
+    "brokerAddress": "localhost:9092,localhost:9093",
```
yvanoers, line 25: I have no experience with Kafka, but I'm guessing having two brokers on the same machine is somewhat odd. If so, can you change the example to something more common?

yvanoers (commented):
Nice, thanks! I have a couple of questions, though.

builtin/bins/dkron-executor-kafka/kafka.go
Espina2: What is your suggestion for that cases? I had that doubt to but I think failing to deliver a message because one of the many addresses passed is wrong it will be worst. This still applies to the single broker, if we provide an extra space it will have the same problem. 🤔 

Maybe pointing to [here](https://jaceklaskowski.gitbooks.io/apache-kafka/content/kafka-properties-bootstrap-servers.html) in the docs will be enough?

website/content/usage/executors/kafka.md
Espina2: Addressed. Put the same example as the official Kafka docs.

vcastellm (dismissed):
LGTM

Espina2:
> LGTM

This are failing tough

vcastellm:
> > LGTM
> 
> This are failing tough

I'm checking the CI

Espina2:
Let me know if there is anything I can help

vcastellm:
SSL tests certs updated, rebase with master and it should work @Espina2

Espina2:
@vcastellm done :)

vcastellm (approved):
LGTM

PR #1109, opened 2022-05-11
Leadership transfer

vcastellm:
* Gracefully transfer leadership
* Fix "cron is already configured"


fopina:
this will allow for even cleaner rolling updates right? I assume transferring leadership is better than removing the leader and triggering/waiting a re-election

PR #1141, opened 2022-06-22
all md files in root path translated to spanish

johnfelipe:
Signed-off-by: felipe <ingenierofelipeurrego@gmail.com>


vcastellm:
I'm a native spanish speaker too @johnfelipe, but what's the point of having these documents translated. Project development language is English and I don't want to maintain spanish translation.

Issue #866, opened 2020-12-11
Schedule jitter needed...


alexanderfefelov:
...like this

> The `RANDOM_DELAY` variable allows delaying job startups by random amount of minutes with upper limit specified by the variable.

or this

> `-j jitter`
> Enable time jitter.  Prior to executing commands, cron will sleep a random number of seconds in the range from 0 to `jitter`.

because

> This option can help to smooth down system load spikes during moments when a lot of jobs are likely to start at once, e.g., at the beginning of the first minute of each hour.

victorcoder:
Thanks for the suggestion, I like it

yvanoers:
Upon seeing my thoughts were: If it is known that a lot of jobs start at once and it is OK for them to not run exactly at the scheduled time, then why not schedule them at different times in the first place?

I'm basically looking for the justification that this should be the responsibility of a scheduler.

victorcoder:
Good question and this is what I do, and I like it because it will take less mental effort to drift a lot of jobs by hand and will specially be useful when working with multiple people/teams creating jobs

yvanoers:
I see.
I'm not a fan of using randomization to accomplish load spreading. While that would usually work fine, there still is a chance of it doing everything around the same time, potentially causing problems that are hard to track down.

A more deterministic approach is a lot less trivial to implement, though.
Hmmm. I'm going to think about this some more.

yvanoers:
I came across an implementation that may be useful: The 'H' as used in jenkins.
See the H explanation in the [Cron article on Wikipedia](https://en.wikipedia.org/wiki/Cron). See its [referenced source](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly) for additional info on the syntax and possibilities in Jenkins.

Relevant bit from Wikipedia:

> 'H' is used in the Jenkins continuous integration system to indicate that a "hashed" value is substituted. Thus instead of a fixed number such as '20 * * * *' which means at 20 minutes after the hour every hour, 'H * * * *' indicates that the task is performed every hour at an unspecified but invariant time for each task. This allows spreading out tasks over time, rather than having all of them start at the same time and compete for resources.

I think that would be a great solution to this problem. What are your thought?

alexanderfefelov:
> What are your thought?

Sounds very interesting. Thank you.

yvanoers:
@victorcoder What do you think?
[...]

PR #1260, opened 2023-01-25
add support for hash scheduling

fopina:
This PR adds support for Jenkins-style hash schedule spec such as
```
0 0 H * * *
```

`H` is replaced by something derived from job name and moduled to its position within the spec (such as `H mod 60` for minutes but `H mod 24` for hour).

This allows for easily scheduling many daily jobs with same cron spec while still mildly balancing them throughout the day.

I tried to do it directly in the dkron `Scheduler` but it gets tricky due to the need of a second parameter that is not part of cron.Scheduler interface.

Also, the `hash` itself is actually just summing up the character values for simplification. It could be improved with an actual hashing method with less collisions and better distribution but in the end it would still use modullus 7, 12, 24, 60 causing many collisions.  
As this is not for security but only _mildly load balancing_, I left as is.

Let me have feedback!

### DIFF
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/distribworks/dkron/v4/extcron"
@@ -303,10 +305,44 @@ func (j *Job) GetTimeLocation() *time.Location {
 	return loc
 }
 
+// scheduleHash replaces H in the cron spec by a value derived from job Name
+// such as "0 0 H * * *"
+func (j *Job) scheduleHash() string {
+	spec := j.Schedule
+	if strings.Contains(spec, "H") && strings.Count(strings.TrimSpace(spec), " ") == 5 {
+		h := 0
+		for _, c := range j.Name {
+			h += int(c)
+		}
+		parts := strings.Split(spec, " ")
+		for index, part := range parts {
+			if strings.Contains(part, "H") {
+				// mods taken in accordance with https://dkron.io/docs/usage/cron-spec/#cron-expression-format
+				ph := h
+				switch index {
+				case 2:
+					ph %= 24
+				case 3:
+					ph = (ph % 31) + 1
+				case 4:
+					ph = (ph % 12) + 1
+				case 5:
+					ph %= 7
+				default:
+					ph %= 60
+				}
+				parts[index] = strings.ReplaceAll(part, "H", strconv.Itoa(ph))
+			}
+		}
+		return strings.Join(parts, " ")
+	}
+	return spec
+}
+
 // GetNext returns the job's next schedule from now
 func (j *Job) GetNext() (time.Time, error) {
 	if j.Schedule != "" {
-		s, err := extcron.Parse(j.Schedule)
+		s, err := extcron.Parse(j.scheduleHash())
 		if err != nil {
 			return time.Time{}, err
 		}
@@ -367,7 +403,7 @@ func (j *Job) Validate() error {
 
 	// Validate schedule, allow empty schedule if parent job set.
 	if j.Schedule != "" || j.ParentJob == "" {
-		if _, err := extcron.Parse(j.Schedule); err != nil {
+		if _, err := extcron.Parse(j.scheduleHash()); err != nil {
 			return fmt.Errorf("%s: %s", ErrScheduleParse.Error(), err)
 		}
 	}
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -33,6 +33,9 @@ const (
 	ConcurrencyAllow = "allow"
 	// ConcurrencyForbid forbids a job from executing concurrency.
 	ConcurrencyForbid = "forbid"
+
+	// HashSymbol is the "magic" character used in scheduled to be replaced with a value based on job name
+	HashSymbol = "~"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


yvanoers:
I'll try and have a closer look this weekend, but at the very least I'd like to see some accompanying unit tests.
Also, please have a look at the discussion in #866 in case you haven't yet.

fopina:
Hi @yvanoers, I had not seen that PR no, but (as mentioned in the original description) Jenkins was my motivation indeed.

This is actually a PR I've had in my ToDo list since I started using dkron 2 years ago, as I used `H` all the time in Jenkins and missed it.

Was there any highlights/caution you wanted me to notice there? My understanding, now that I read it, is that everyone agreed with the approach.

Definitely missing unit tests, I'll take care of those

Note: I'm already using this in my fork, so far so good!

yvanoers:
Hi @fopina !
The noteworthy bit of the feature request is the intent to have this functionality added to the cron library - that's where it should belong. There ought to come a PR having the same behavior as this PR for that project, so we can seemlessly move to a new version of cron when (and if) it'll get merged.
I can't make you do that, of course, but it would be much appreciated.

fopina:
ah got it @yvanoers !

I think it would make sense to make it available everywhere though… cron library cares about schedule nothing else.

There’s not enough entropy in the schedule string to use it as the hash meaning a new non-sense parameter would be added to the library method just so it would do the “strings.replace” anyone can easily do before calling it.

So even though I think it would make sense, I don’t think it does in the end 😄 

but happy to agree to disagree and closing this PR!

dkron/job.go
```diff
@@ -303,10 +305,44 @@ func (j *Job) GetTimeLocation() *time.Location {
...
+			h += int(c)
+		}
+		parts := strings.Split(spec, " ")
+		for index, part := range parts {
+			if strings.Contains(part, "H") {
+				// mods taken in accordance with https://dkron.io/docs/usage/cron-spec/#cron-expression-format
+				ph := h
+				switch index {
+				case 2:
+					ph %= 24
```
yvanoers, line 324: Not all days are 24 hours long. Not sure how the Jenkins scheduler handled that, but it is worth thinking about and documenting.

dkron/job.go
```diff
@@ -303,10 +305,44 @@ func (j *Job) GetTimeLocation() *time.Location {
...
+				case 3:
+					ph = (ph % 31) + 1
```
yvanoers, line 326: Not all months are 31 days long.

yvanoers:
> ... and closing this PR!

There are no high expectations of this getting included in the cron scheduler.
And we do want to support it in Dkron.
So no need to close this PR prematurely.

Will finish code review hereafter.

yvanoers (changes requested):
I'm thinking that doing this implementation correctly while preventing any undesired/unexpected behavior may need to be done by parsing the schedule in greater detail.
Which is why the cron scheduler itself is arguably the best place for this functionality. Otherwise it may be necessary to parse the schedule twice, which feels wasteful to me.

I'm happy to be proven wrong though :)

Also: We're not against including a modified version of the entire cron scheduler in Dkron. It's been so before. Ideally we wouldn't have to, though, but that is probably not feasible in the short term.

fopina:
Thank for taking the time for explaining @yvanoers , I do understand your point much better now.

Still I'm not sure how to introduce the hashable parameter without making it look nonsense. But I'll definitely take a look first then!

yvanoers:
> Maybe something more unique that we wouldn't have to parse schedule in greater detail?

That may be worth exploring.
'H' is an unofficial addition anyway, there's no absolute need to adhere to it.
'~' perhaps?

cobolbaby:
Should we consider implementing a new nodeSelector to support hash scheduling strategies? Then introduce a new configuration item to switch the scheduling strategy? The default nodeSelector strategy is random dispatch. 

https://github.com/distribworks/dkron/blob/f114610b9002d21a1d959be92e356cb02cda5991/dkron/run.go#L33

https://github.com/distribworks/dkron/blob/f114610b9002d21a1d959be92e356cb02cda5991/dkron/agent.go#L757-L760

dkron/job.go
```diff
@@ -303,10 +305,44 @@ func (j *Job) GetTimeLocation() *time.Location {
 	return loc
 }
 
+// scheduleHash replaces H in the cron spec by a value derived from job Name
+// such as "0 0 H * * *"
+func (j *Job) scheduleHash() string {
+	spec := j.Schedule
+	if strings.Contains(spec, "H") && strings.Count(strings.TrimSpace(spec), " ") == 5 {
```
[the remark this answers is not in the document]
fopina: refactored a bit to handle `TZ` `CRON_TZ` and `@...`, thanks for the heads up!

dkron/job.go
```diff
@@ -303,10 +305,44 @@ func (j *Job) GetTimeLocation() *time.Location {
...
+		h := 0
```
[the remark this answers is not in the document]
fopina: done, hope those look clearer now

dkron/job.go
```diff
@@ -303,10 +305,44 @@ func (j *Job) GetTimeLocation() *time.Location {
...
+		for _, c := range j.Name {
+			h += int(c)
+		}
+		parts := strings.Split(spec, " ")
+		for index, part := range parts {
+			if strings.Contains(part, "H") {
```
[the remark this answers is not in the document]
fopina: It is intended yes, not because of `*/H` but for `H/5` (every 5min starting at `H`).
Not sure if `*/H` would be of any use, but I don't think it's worth actively excluding it as `*/55` (every 55minutes) is still valid cron expression 🤷 
What do you think?

dkron/job.go
fopina: Not sure I understood? I'm probably missing when are days not 24h long.
Also, that doesn't seem to be an "hash" issue since most people setting up jobs to run at specific hour every day, will choose an hour out of 24...
The reference for these mods are the official dkron documentation for cron expressions, hour part is `0-23`

dkron/job.go
fopina: While for hours I didn't understand your point, I do in this one 😄 
Updated max to `28` 🙇 
Curious cron library would handle choosing day 31 for a month of 30, as anyone can specify that in a spec, will it skip a month?

fopina:
> '~' perhaps?

@yvanoers thank you for all the feedback. I've also updated `H` to `~` while going through the rest of the review comments.

Hope it looks better now!

Sadly, I couldn't reply to review comments within a single review (sorry for the spam), because of the force push I had to do when switching the PR to v4, but it felt like a required step as well ⚡

dkron/job.go
yvanoers: Actually, the day not being 24h long is a bit besides the point - sorry. 

My point was daylight savings time:
When daylight savings time kicks in, for timezones that have that, 2 AM is skipped over: 1:59AM -> 3:00AM. If something is scheduled to jitter around 3AM, it may decide to pick 2AM on a day that does not have 2AM.
The other way around, when coming back from daylight savings time, you have 2AM twice: 2:59AM -> 2:00AM.

This is why it is generally a good idea to run scheduling servers on UTC time. Running a scheduler with a timezone that can jump back and forth in time is risky. Hence, worth thinking about and documenting.

fopina: Oh nice point on DST! Good stuff to keep in mind for many things!
Personally I have every server in UTC; not just cronjobs, because I never know which apps write logs with UTC or localtime and then depending on the visualization tool, again the same issue, so I'm totally onboard with the UTC recommendation.

That being said, I don't think I agree that it should be worth thinking and documenting "within this PR" but rather at the scheduler level (maybe not even dkron, but the cron lib).
This PR only adds a special character that will resolve to number. And it will resolve for valid values of the field where it's used.
If there are values that might result in bad behavior, it's still the same as if the user would choose that specific value himself (instead of using `H`/`~`).
I agree that we should avoid cases that make sense (like you mentioned for months, hope you agree with using `28` as limit), but for dealing with DST and the issues you mention, "once a day" schedule will be subject to them, regardless of using `0 0 H * * *` or `0 0 3 * * *`.
So I think it's a valid caution/recommendation to note somewhere, but not in the `scheduleHash` specifically, as the risk is not introduced by using it.
What do you think?

fopina:
@yvanoers any chance for review? :)

fopina:
is it worth the effort to rebase this PR with v4?

yvanoers:
> is it worth the effort to rebase this PR with v4?

Isn't it already targeted at v4 ?

fopina:
No, I don’t think there was a v4 when this branch was created

yvanoers:
Well, you changed the base from main to 4.x 7 months ago and did a couple force pushes. So if you didn't rebase to the 4.x branch back then already, I would think is does make sense to rebase it.
The changeset is still 3 files, though, so I'm not worried it will not properly merge if it isn't rebased.

vcastellm:
Hey this somehow keeps escaping my radar sorry, I'll take a look and get back to you, it looks like it'll merge right away.

fopina:
@yvanoers you're right, i probably rebased and can’t remember

@vcastellm looks like it indeed, thanks!

vcastellm (commented):
LGTM but can you add docs in https://github.com/distribworks/dkron/blob/4.x/website/docs/usage/cron-spec.md ? thx

fopina:
right, sorry for missing that @vcastellm, done now 👍

website/docs/usage/cron-spec.md
```diff
@@ -51,6 +51,10 @@ Question mark ( ? )
 Question mark may be used instead of '*' for leaving either day-of-month or
 day-of-week blank.
 
+Tilde ( ~ )
+
+Tilde will be replaced by a numeric value valid for the range where it is used. It allows periodically scheduled tasks to produce even load on the system. For example, scheduling multiple hourly jobs to "0 H * * * *" rather than "0 0 * * * *" will run the jobs at different minutes of every hour. It can be thought of as a random value over a range, but it actually is a hash of the job name, not a random function, so that the value remains stable for any given job.
```
vcastellm, line 56: Should this be `"0 ~ * * * *"`?

fopina: 🤦 inspired myself in jenkins docs and forgot to replace!

PR #1279, opened 2023-02-27
support custom Host header in webhook-headers and pre-webhook-headers

fopina:
Apparently Go http Client overrides `Host` in the headers. One has to use `request.Host` to set it.

This change allow custom `Host` header to be used like any other.


vcastellm (approved):
LGTM

PR #1312, opened 2023-04-07
Add logrus Hook to write logs to stderr and stdout based on log level

sandyydk:
**Issue**

- Currently when we use DKRON as deployment in GCP k8s, the logs emitted by DKRON are by default captured as error logs by GCP Stackdriver. This is because by default logrus writes all logs to stderr. This gets misclassified by Stackdriver.

**Solution**

- Add a hook to logrus so that based on log levels they are written appropriately to the different output channels (stdout and stderr).

### DIFF
--- a/cmd/dkron.go
+++ b/cmd/dkron.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/distribworks/dkron/v3/dkron"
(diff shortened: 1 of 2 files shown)


vcastellm (approved):
LGTM thanks!

PR #1357, opened 2023-07-05
Dockerfile加入goproxy

zy410692:
docker-compose up -d is too slow.


yvanoers (changes requested):
We are not going to set a Go proxy specific to a certain region directly in the dockerfile.
And that is besides any privacy and security concerns this may present.

I would imagine it is possible to set GOPROXY from the docker-compose command line somehow.

PR #1384, opened 2023-09-13
add TLS options for kafka executor plugin

hosseinsadeghi-cpi:
This PR adds TLS configurations for Kafka executor to enable users to communicate with Kafka clusters protected with encryption.


yvanoers (approved):
LGTM

vcastellm (approved):
LGTM thanks for the contrib!

PR #1426, opened 2023-12-06
Update release.yml

timofey-kurakin-m10:


yvanoers:
What is the purpose of this change?

PR #1434, opened 2023-12-09
Fix shell command

vcastellm:
In upcoming release we're going to drop window support.

### DIFF
--- a/builtin/bins/dkron-executor-shell/shell.go
+++ b/builtin/bins/dkron-executor-shell/shell.go
@@ -1,17 +1,10 @@
 package main
 
 import (
-	"encoding/base64"
 	"errors"
-	"fmt"
-	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


djalan:
> In upcoming release we're going to drop window support.

What is the latest version of Dkron that can run under Windows?

Or was your comment only related to shell executor?

vcastellm:
It's actually related to the shell executor only but is Dkron useful without the shell executor? to certain degree only.

To my knowledge it's still usable in Windows, but we don't test it.

Are you using it in Windows @djalan ?

djalan:
Hello @vcastellm

Yes I am using it in Windows and exclusively with the shell executor.

I had to go back from 4.0.2 to 3.7 because I am getting network errors almost all the time when I try to launch a job. I turned off the firewall in Windows, but it did not help.

PR #1500, opened 2024-03-19
Minor improvement in RabbitMQ executor

ivan-kripakov-m10:
Provide possibility to decide whether to create queues when using RabbitMQ executor and possibility to choose queue's configuration (`durable`, `auto_delete`,  `exclusive`) if user decides to create the queue via plugin. 
Also allows users to customize message publishing using `messageId`, `contentType` and `delivery_mode` options.

### DIFF
--- a/builtin/bins/dkron-executor-rabbitmq/rabbitmq.go
+++ b/builtin/bins/dkron-executor-rabbitmq/rabbitmq.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/base64"
 	"errors"
+	"strconv"
 
 	dkplugin "github.com/distribworks/dkron/v4/plugin"
 	dktypes "github.com/distribworks/dkron/v4/types"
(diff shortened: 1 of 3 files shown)


builtin/bins/dkron-executor-rabbitmq/rabbitmq.go
```diff
@@ -32,64 +40,117 @@ func (r *RabbitMQ) Execute(args *dktypes.ExecuteRequest, cb dkplugin.StatusHelpe
...
+
+	// publish message
+	if err = publish(cfg, ch); err != nil {
 		return nil, err
 	}
+	return nil, nil
+}
+
+func createQueueIfNecessary(cfg map[string]string, queue string, ch *amqp.Channel) error {
+	if val, ok := cfg["queue.create"]; !ok || (ok && val == "false") {
```
vcastellm, line 98: Should this query rabbit in order to check if the queue already exists?

vcastellm (approved):
LGTM aside from comment

builtin/bins/dkron-executor-rabbitmq/rabbitmq.go
ivan-kripakov-m10: I think we don't need to check if the queue exists here. If the user chooses not to create the queue via the plugin, we expect him to create it manually. 
If the user doesn't declare the queue, message publishing will fail with error in logs or a failed execution in UI.

PR #1543, opened 2024-06-19
Docs: Fix Prometheus metrics port

artemik:
Seems like port 6080 is a typo in the docs, and it should be instead the default UI port 8080.

On a fresh Dkron installation I also don't see anything listening on 6080 port.

https://dkron.io/docs/usage/metrics


mrhackcz:
Its actually listening on port `9422`

mrhackcz (approved):
LGTM

vcastellm (approved):

Issue #1570, opened 2024-08-28
Shell Executor Prometheus metrics: calculating % of failed jobs should be easy


atmosx:
**Is your feature request related to a problem? Please describe.**

I’m trying to create a Grafana dashboard based on the shell executor metrics. One of the graphs is meant to calculate the percentage of failed jobs over a specific time frame or from the start of tracking.

The `dkron_job_execution_done_count` metric is a counter. A common approach is to filter by a specific label versus all entries of the same metric. This can be achieved by adding `exit_code` as a label to `dkron_job_execution_done_count`. Since it’s a low cardinality metric, adding this label should not pose any issues.

**Describe the solution you'd like**
Add `exit_code` to `dkron_job_execution_done_count`.

**Describe alternatives you've considered**
None.

vcastellm:
Already fixed

PR #1576, opened 2024-09-09
feat: add `exit_code` label to `execution_done_count`

atmosx:
The `exit_code` label adds useful information to the `execution_done_count` metric. The exit codes are usually `0`, `1` or `127`, this is a low cardinality metric.

This label will allow us use promQL to perform standard operations like calculate the number of failed vs total jobs.

### DIFF
--- a/builtin/bins/dkron-executor-shell/prometheus.go
+++ b/builtin/bins/dkron-executor-shell/prometheus.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"log"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


atmosx:
Fixes #1570

vcastellm (approved):
LGTM Thanks for the contrib!

atmosx:
@vcastellm sorry, made a mistake and pushed two commits to this repo instead, dropped the two new commits and force-pushed.

vcastellm (approved):

PR #1608, opened 2024-11-05
use *types.Execution to avoid sync.Mutex copy

BrainGithub:
1. Use *types.Execution to avoid sync.Mutex copy, with slightly interface change also.
2. Upgrade golang image from 1.22 to 1.23.
3. Upgrade deprecated file io calls.
    - `ioutil.WriteFile` -> `os.WriteFile`
    - `ioutil.ReadAll` -> `io.ReadAll`


BrainGithub:
@vcastellm, all check passed, please help review, thanks.

PR #1625, opened 2024-12-13
Prevent calling output on an undefined record

djalan:
Hello

I upgraded to Dkron 4.0.0 for Windows AMD64 and the UI crashes for Busy job when I try to expand the output. It used to work on version 3.7 for the same shell command line.

https://github.com/distribworks/dkron/blob/e232733f0d13d22e13031e6f3ebf2a34d3e33a5a/ui/src/executions/BusyList.tsx#L5

A quick fix would be

```tsx
record?.output
```

image


vcastellm (approved):
thanks for the contribution @djalan LGTM

PR #1652, opened 2025-01-14
Fix help message for remove-peer raft command

morgan-atproperties:
The current message was duplicated from the list-peers raft command.


yvanoers (approved):

vcastellm (approved):

vcastellm:
thx @morgan-atproperties !

PR #1680, opened 2025-03-07
fix: RabbitMQ executor configuration/payload validation

xBlaz3kx:
Seems like there was a discrepancy in the RabbitMQ executor payload validation and actual implementation for publishing base64 encoded bodies.

Validation prevented both the unencoded body and base64 encoded body to be set, but in when publishing, it checked the wrong attribute. This could allow both the `message.body` and `"message.base64Body"` to be set and still publish the message.

Also: 
- Added logs to RabbitMQ executor
- Added test cases for configuration validation


vcastellm (approved):
LGTM thanks!

PR #1682, opened 2025-03-07
chore: PR template

xBlaz3kx:
I went ahead and added a PR template I use regularly to mainstream the PRs, so they should be a bit more readable and include the purpose and other information for a PR.


vcastellm (changes requested):
Even though is good to have a PR template I want the PRs descriptions to stay lean and small, I usually hate too verbose PR templates forcing to fill multiple checkboxes that doesn't apply in most of the cases.

So my suggestion here would be to remove the checklists or reduce them to the minimum.

vcastellm (approved):

PR #1685, opened 2025-03-07
fix: Context timeout for Google Cloud PubSub

xBlaz3kx:
Added context timeout to prevent indefinite job execution for GC PubSub. 
The PubSub client however has a keepalive of 5 minutes set, but that is way too much imo.


vcastellm (approved):
LGTM thx!

PR #1703, opened 2025-03-18
Feature(ci): Autogen API client on spec change

xBlaz3kx:
## Proposed changes

- Added workflow for validating OpenAPI spec
- Added autogeneration of the OpenAPI client when the spec changes

(Solves ##1701)

## Types of changes


xBlaz3kx:
Note: New workflow wasnt run because OpenAPI spec was not changed. If you want to try out the workflow, I suggest checking out my branch and running the action using `nektos/act` tool or you can take this [PR](https://github.com/xBlaz3kx/dkron/pull/2) as a reference. 

Sidenote 2: OpenAPI validation is failing, not sure why.

.github/workflows/autogen.yaml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          go-version: 1.23.0
+
+      - name: Install Oapi Codegen and generate client
+        run: |
+          go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
+          oapi-codegen -package client website/static/openapi/openapi.yaml > ./client/client.go
+
+      - name: Push the generated code
+        uses: stefanzweifel/git-auto-commit-action@v5
```
yvanoers, line 57: I'm OK with the autocommit when triggered from a PR, but not on main.
Committing/pushing to main shouldn't happen, but if it does I'd like to see it make a PR instead.
Same when triggered manually.

yvanoers (changes requested):

.github/workflows/autogen.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.23.0
+
+      - name: Install Oapi Codegen and generate client
+        run: |
+          go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
+          oapi-codegen -package client website/static/openapi/openapi.yaml > ./client/client.go
```
yvanoers, line 54: Could you call `make client` here instead of calling oapi-codegen directly? Saves us maintenance of the same line of code.

yvanoers (changes requested):
Nice, thanks! One more thing, though.

yvanoers (approved):

.github/workflows/autogen.yaml
```diff
@@ -0,0 +1,72 @@
...
+        if : github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "[Autogen] Generated client from OpenAPI spec"
+          title: "[Autogen] OpenAPI client update"
+          body: "OpenAPI client was (re)generated from the latest OpenAPI spec."
+          branch: "openapi/client/autogen-${{ github.sha }}"
+          base: main
+          labels: openapi-client, autogen
```
vcastellm, line 72:
```suggestion
          labels: openapi-client, autogen

```

PR #1705, opened 2025-03-19
Chore: Improving docker compose

xBlaz3kx:
## Proposed changes

- Added a testing-ready version of Dkron that uses Dockerhub image
- Updated docs to reflect the docker compose changes
- Improved docs' quickstart

## Types of changes

- [x] Documentation Update (if none of the other choices apply)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,29 +1,31 @@
----
-version: '2'
 services:
+  # Run a leader node
   dkron:
-    build: .
+    image: dkron/dkron:4.0.4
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,29 +1,31 @@
----
-version: '2'
 services:
+  # Run a leader node
   dkron:
-    build: .
+    image: dkron/dkron:4.0.4
```
vcastellm, line 4: Use `latest` here, I don't want to keep updating this version and, more importantly, this compose should always work with latest.

PR #1706, opened 2025-03-19
Fix: Use log instead of fmt

xBlaz3kx:
## Proposed changes

- Replaced fmt usages in agent with logs to keep thing consistent

## Types of changes

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -89,12 +89,12 @@ WAIT:
 	case s := <-signalCh:
 		sig = s
 	case err := <-agent.RetryJoinCh():
-		fmt.Println("[ERR] agent: Retry join failed: ", err)
+		log.WithError(err).Error("agent: Retry join failed")
 		return 1
 	case <-ShutdownCh:
 		sig = os.Interrupt
 	}
-	fmt.Printf("Caught signal: %v", sig)
+	log.Infof("Caught signal: %v", sig)
 
 	// Check if this is a SIGHUP
 	if sig == syscall.SIGHUP {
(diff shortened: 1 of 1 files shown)


cmd/version.go
```diff
@@ -14,10 +13,10 @@ var versionCmd = &cobra.Command{
 	Short: "Show version",
 	Long:  `Show the version`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Name: %s\n", dkron.Name)
-		fmt.Printf("Version: %s\n", dkron.Version)
-		fmt.Printf("Codename: %s\n", dkron.Codename)
-		fmt.Printf("Agent Protocol: %d (Understands back to: %d)\n",
+		log.Infof("Name: %s\n", dkron.Name)
```
vcastellm, line 16: This is showing the version, it's not logging, so no motivation to use log.

cmd/version.go
```diff
@@ -14,10 +13,10 @@ var versionCmd = &cobra.Command{
...
+		log.Infof("Version: %s\n", dkron.Version)
+		log.Infof("Codename: %s\n", dkron.Codename)
+		log.Infof("Agent Protocol: %d (Understands back to: %d)\n",
```
vcastellm, line 19: This is showing the version, it's not logging, so no motivation to use log.

vcastellm (changes requested):
View comments

cmd/version.go
xBlaz3kx: Logrus will also output the version into the console/stdout. This was changed just to make things a bit more consistent

vcastellm: This is the issue
image

I don't want the logrus log prefix in the version command, it's polluting the output in this case.

xBlaz3kx: Alright, then I'll revert

PR #1707, opened 2025-03-21
Feature: Docker image healthcheck

xBlaz3kx:
## Proposed changes

- Added healthcheck to Dockerfile

## Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,7 @@ COPY . .
 
 RUN go install ./...
 
+HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=5 \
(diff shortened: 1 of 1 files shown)


vcastellm:
This Dockerfile is just for development, can you elaborate on why do you need this?

xBlaz3kx:
Well in general its useful to know if the container is healthy or not. For example when running the raft cluster, one of the nodes could not be connected to the cluster, but in general that wouldnt terminate the container. 

I guess the healthcheck does reflect that? 

I'm also quite unsure why, by default, the docker entrypoint/cmd is just `dkron` command and not `dkron agent --server` ?
It would make it easier to run a cluster out of the box

PR #1761, opened 2025-06-24
Fix: keep the existing dependent_jobs value on updating an existing job

indeedhat:
## Proposed changes

Instructs the store to include existing dependent_jobs on job update.  
Currently the dependent_jobs slice gets reset to nil on updating an existing job, this breaks the job chaining feature.

## Types of changes

### DIFF
--- a/dkron/api_test.go
+++ b/dkron/api_test.go
@@ -141,6 +141,63 @@ func TestAPIJobCreateUpdateParentJob_NoParent(t *testing.T) {
 	assert.Contains(t, string(errJSON)+"\n", string(body))
 }
 
+func TestAPIJobCreateUpdateParentJob_KeepDependents(t *testing.T) {
+	port := "8101"
+	baseURL := fmt.Sprintf("http://localhost:%s/v1", port)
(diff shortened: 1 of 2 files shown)


indeedhat:
That is a fair point, ill work out how to do that.

indeedhat:
Upon further investigation im not sure this is an issue.

If im reading it correctly in `Store.SetJob` store.go:122 and more specifically the `DB.Update` callback it attempts to load the existing job and only if it finds one will overwrite existing values including the `DependentJobs` before save.

I could certainly update the applySetJob to look something like:
```go
func (d *dkronFSM) applySetJob(ctx context.Context, buf []byte) interface{} {
	var pj dkronpb.Job
	if err := proto.Unmarshal(buf, &pj); err != nil {
		return err
	}
	_, err := d.store.GetJob(ctx, pj.Name, nil)
	jobExists := err != nil
	job := NewJobFromProto(&pj, d.logger)
	if err := d.store.SetJob(ctx, job, jobExists); err != nil {
		return err
	}
	return nil
}
``` 

But i just want to make sure im not duplicating work before i do.

PR #1770, opened 2025-07-20
feat: improve openapi spec and http responses

xBlaz3kx:
## Proposed changes

- Fix OpenAPI spec to be up to date with the implementation 
- Fixed error case when get jobs cannot be executed to return an internall error

## Types of changes

- [x] Documentation Update (if none of the other choices apply)


xBlaz3kx:
Would be nice if #1773 was merged before this one just to split up the PR a bit nicer.

xBlaz3kx:
@vcastellm mind reviewing?

PR #1785, opened 2025-09-01
feat: Expose key operational metrics for Prometheus

ashleefr:
This pull request introduces native Prometheus metrics to improve the observability of a Dkron agent. While Dkron uses go-metrics internally, it lacks a direct, out-of-the-box way to expose crucial operational data in the Prometheus format, which has become a standard for cloud-native monitoring.

This change adds a new prom_metrics.go file and instruments key parts of the application to expose the following metrics:

*   dkron_agent_event_received_total{event="..."}: A counter for all events received by the agent's event loop, labeled by event type. This helps in understanding the agent's activity and message flow.
*   dkron_job_executions_succeeded_total{job_name="..."}: A counter for successfully completed job executions, labeled by the job name.
*   dkron_job_executions_failed_total{job_name="..."}: A counter for failed job executions, labeled by the job name.

These metrics are essential for:
- Building dashboards to monitor the health of the Dkron cluster.
- Setting up alerts on important events, such as a high rate of job failures for a specific job.
- Gaining better insight into the system's behavior over time.

The init() function in prom_metrics.go is used to pre-register some label values for agentEventReceivedTotal, ensuring that the time series exist even before the first event of that type occurs. This is a common practice to avoid missing metrics at startup.

### DIFF
--- a/dkron/agent.go
+++ b/dkron/agent.go
@@ -710,6 +710,8 @@ func (a *Agent) eventLoop() {
 		case e := <-a.eventCh:
 			a.logger.WithField("event", e.String()).Info("agent: Received event")
 			metrics.IncrCounter([]string{"agent", "event_received", e.String()}, 1)
+			// Prometheus mirror
+			agentEventReceivedTotal.WithLabelValues(e.String()).Inc()
 
 			// Log all member events
 			if me, ok := e.(serf.MemberEvent); ok {
--- a/dkron/grpc.go
+++ b/dkron/grpc.go
@@ -8,9 +8,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/armon/go-metrics"
 	"github.com/distribworks/dkron/v4/plugin"
 	"github.com/distribworks/dkron/v4/types"
+	"github.com/hashicorp/go-metrics"
 	"github.com/hashicorp/raft"
 	"github.com/hashicorp/serf/serf"
 	"github.com/sirupsen/logrus"
@@ -210,6 +210,12 @@ func (grpcs *GRPCServer) ExecutionDone(ctx context.Context, execDoneReq *types.E
 	// If the execution failed, retry it until retries limit (default: don't retry)
 	// Don't retry if the status is unknown
 	execution := NewExecutionFromProto(&pbex)
+	// Prometheus job execution counters
+	if execution.Success {
+		jobExecutionsSucceededTotal.WithLabelValues(job.Name).Inc()
+	} else {
+		jobExecutionsFailedTotal.WithLabelValues(job.Name).Inc()
+	}
 	if !execution.Success &&
 		uint(execution.Attempt) < job.Retries+1 &&
 		!strings.HasPrefix(execution.Output, ErrBrokenStream.Error()) {
--- a/dkron/prom_metrics.go
+++ b/dkron/prom_metrics.go
@@ -1,0 +1,38 @@
+package dkron
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	// dkron_agent_event_received_total{event="..."}
+	agentEventReceivedTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "dkron",
+		Subsystem: "agent",
+		Name:      "event_received_total",
+		Help:      "Count of events received by the agent",
+	}, []string{"event"})
+
+	// dkron_job_executions_failed_total{job_name="..."}
+	jobExecutionsFailedTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "dkron",
+		Subsystem: "job",
+		Name:      "executions_failed_total",
+		Help:      "Total number of failed job executions",
+	}, []string{"job_name"})
+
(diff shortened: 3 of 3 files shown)


vcastellm:
Thanks for the contribution @ashleefr but I don't understand the purpose of the PR, Dkron already have Prometheus metrics out-of-the-box https://dkron.io/docs/usage/metrics#prometheus, can you specify what are you lacking in your use case?

ashleefr:
> Thanks for the contribution @ashleefr but I don't understand the purpose of the PR, Dkron already have Prometheus metrics out-of-the-box https://dkron.io/docs/usage/metrics#prometheus, can you specify what are you lacking in your use case?

Thank you for the quick response! You're absolutely right that Dkron has Prometheus metrics out-of-the-box. However, after investigating the current implementation, I found that the metrics listed in the documentation (like `dkron.agent.event_received.query_execution_done`) are not actually exposed on the `/metrics` endpoint in a format that's useful for monitoring.

The current go-metrics integration with Prometheus sink doesn't seem to be properly exposing the agent event metrics and job execution outcomes that are mentioned in the docs. When I checked the `/metrics` endpoint, I only saw basic go-metrics internal counters, but not the specific agent events or job success/failure counts that would be essential for alerting.

My PR specifically addresses this gap by:

1. Adding native Prometheus counters for agent events (`dkron_agent_event_received_total`) that actually track the events mentioned in the docs
2. Adding job execution success/failure counters (`dkron_job_executions_succeeded_total`, `dkron_job_executions_failed_total`) that are crucial for monitoring job health
3. Ensuring these metrics are properly incremented in the relevant code paths

I've kept this PR minimal with just a few key metrics to demonstrate the issue, but ideally I'd like to see all the metrics mentioned in the documentation (https://dkron.io/docs/usage/metrics#prometheus) actually working and properly exposed. The current implementation seems to have a gap between what's documented and what's actually available.

Could you help me understand if there's a configuration I'm missing, or if the current Prometheus integration is indeed not exposing these specific metrics? I'd be happy to adjust the PR based on your feedback or expand it to cover all the documented metrics if that would be more helpful.

I can provide a test setup with docker-compose if that would help demonstrate the issue I'm trying to solve.

vcastellm:
> The current go-metrics integration with Prometheus sink doesn't seem to be properly exposing the agent event metrics and job execution outcomes that are mentioned in the docs. When I checked the `/metrics` endpoint, I only saw basic go-metrics internal counters, but not the specific agent events or job success/failure counts that would be essential for alerting.

This was a regression already fixed in https://github.com/distribworks/dkron/pull/1786,

ashleefr:
Hi @vcastellm,

Thanks again for the information. I looked into PR #1786 and saw that you added the `hashicorpmetrics` build tag. I noticed that the `Dockerfile` was not updated in that change, so I modified it locally to include the tag for my testing. After doing so, the base cluster metrics (`raft`, `serf`, etc.) became visible, which is a great step forward.

However, the main issue still persists. The [official documentation](https://dkron.io/docs/usage/metrics#prometheus) mentions specific job metrics like `dkron_job_executions_failed_total`. Even with the build tag fix and with active jobs running, these documented metrics are not being exposed on the `/metrics` endpoint.

So, it seems there are two distinct issues that need to be addressed:

1.  The `Dockerfile` build process is out of sync with the `Makefile`, as it's missing the `hashicorpmetrics` tag.
2.  The application is not instrumented to expose the job execution metrics that are described in the documentation.

Once these issues are resolved, are there any plans for a new release on Docker Hub? It would be very helpful for users to get these fixes in an official image without needing to build from source.

Thanks again for your time and help.

vcastellm:
@ashleefr tnx for checking, yes that was the fix, I will PR to fix the Dockerfile.

Take into account that metrics are not shown in the metrics endpoint until one metric of the specific type is recorded.

Expect the release in the following hours

ashleefr:
Hi @vcastellm,

Thanks for the quick release and for fixing the Dockerfile!

I've tested the new official image. You were right about the Dockerfile fix – the base cluster metrics (`raft`, `serf`, etc.) are now visible, which is great.

Unfortunately, I'm still not seeing the job execution metrics. I followed your advice and ran several successful and failing jobs to trigger the events, but the metrics mentioned in the [official documentation](https://dkron.io/docs/usage/metrics#prometheus) (like `dkron.job.<job name>.success`) are still not being exposed on the `/metrics` endpoint.

It seems there is still a gap between the documented behavior and the current implementation.

Thanks again for your help and the quick turnaround on the Docker fix.

ashleefr:
HI @vcastellm, can you tell me about the question above, please

ashleefr:
Hi @vcastellm

The availability of job execution metrics (like success/failure counters) is a critical requirement for us to use Dkron effectively in a production environment. Without these metrics, we lack the visibility to monitor the health of our most important jobs and cannot create reliable alerts for failures. This kind of detailed, application-level observability is a standard feature in job scheduling systems and is essential for operational stability.

We understand you're likely very busy. Is implementing these documented metrics on your roadmap? If you don't have the bandwidth to work on this at the moment, I would be more than happy to help by contributing a pull request to add the necessary instrumentation.

Thank you for your time and for all your work on Dkron.

vcastellm:
@ashleefr yeam this is why I reopened the issue, we'll take a look to it in the following hours

PR #1838, opened 2025-11-13
feat: startsAt field was added to the Job type

NAlexandrov:
## Proposed changes

This MR adds the field `startsAt` to the `Job` type (`starts_at` for protobuf). If this field has a value and that value is greater than the current date, the job will not be executed until the specified date.

All this will allow you to create delayed tasks that will be executed starting from the specified date.

## Types of changes

### DIFF
--- a/dkron/job.go
+++ b/dkron/job.go
@@ -404,7 +404,7 @@ func (j *Job) isRunnable(logger *logrus.Entry) bool {
 
 	if j.StartsAt.HasValue() && time.Now().Before(j.StartsAt.Get()) {
 		logger.WithField("job", j.Name).
-			Debug("job: Skipping execution because job is not started")
+			Debug("job: Skipping execution because job not due to start yet")
 		return false
 	}
 
--- a/client/client.go
+++ b/client/client.go
@@ -16,11 +16,11 @@ import (
 	"path"
 	"strings"
 
+	"github.com/distribworks/dkron/v4/types"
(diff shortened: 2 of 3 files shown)


dkron/job.go
```diff
@@ -377,9 +396,21 @@ func (j *Job) GetNext() (time.Time, error) {
...
-	if j.Disabled || (j.ExpiresAt.HasValue() && time.Now().After(j.ExpiresAt.Get())) {
+	if j.Disabled {
+		logger.WithField("job", j.Name).
+			Debug("job: Skipping execution because job is disabled")
+		return false
+	}
+
+	if j.StartsAt.HasValue() && time.Now().Before(j.StartsAt.Get()) {
+		logger.WithField("job", j.Name).
+			Debug("job: Skipping execution because job is not started")
```
yvanoers: "job is not started" can be confusing: the job didn't start because it hasn't started...?!
I suggest writing "... job not due to start yet." or something to that effect.

yvanoers (changes requested):
I'd prefer if the changes to the openapi.yaml were in this PR too.
I realize they sit in the front end, but my comment in the discussion was about the other frontend stuff.

yvanoers (approved):

yvanoers:
Alright, looks good to me.
At some point we should add tests for expiresAt and startsAt. But as there looks to be no test for expiresAt either at the moment, I see no harm in making that a future addition.

